PR #4, opened 2020-06-11
Initial frontend

wolfv:
We should probably figure out where to place the `quetz-frontend` folder, and how to name it.


wolfv:
This was superseded by #12

PR #5, opened 2020-06-24
README.md: Force --strict-channel-priority on xtensor install

tom--pollard:
Without having previously set `config --set channel_priority strict` the install for xtensor will resolve to the `conda-forge` package due to the higher version number `xtensor-0.21.5`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Run test upload CLI client:
 
 Install the test package with conda:
 ```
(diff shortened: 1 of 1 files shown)


tom--pollard:
This could also be rectified by pointing at latest release in `download-test-package.sh/test-cli-client.sh` and regenerating the output snippet in the `README`, however that seems like more effort against a moving target.

PR #7, opened 2020-06-25
Perform a URL lookup for the authorize route

benjamb:
Rather than appending a hardcoded route to a URL set in the config.


mariobuikhuizen:
Does this also work behind a reverse proxy or load balancer?

benjamb:
Just tested with nginx, with `proxy_set_header Host $http_host` set in the config and `--proxy-headers` passed to uvicorn this works.

mariobuikhuizen:
Great! Thanks!

PR #8, opened 2020-06-25
Fix wget calls in download script

benjamb:
* Remove duplicate 'wget' arg.
* Remove erroneous '.' arg.


mariobuikhuizen:
Thanks!

PR #9, opened 2020-06-29
feat: quetz cli client in python

mariobuikhuizen:


quetz_client/setup.py
```diff
@@ -0,0 +1,21 @@
...
+import setuptools
+
+
+here = os.path.dirname(os.path.abspath(__file__))
+version_ns = {}
+with open(os.path.join(here, 'quetz_client', '_version.py')) as f:
+    exec(f.read(), {}, version_ns)
+
+
+setuptools.setup(
```
wolfv, line 11: I think instead of creating the `bin/quetz` file, we could / should use a python entrypoint. 

This is how it's done in mamba, for example:

https://github.com/TheSnakePit/mamba/blob/4e7d34802f5e84ff19511f25673d62866b3228fa/setup.py#L155-L159

Also, should we call it `quetz-client`? so that we can still have quetz as the command to start the server?

PR #18, opened 2020-07-08
Redirect to the root after auth/logout

benjamb:
Allow the web application to serve either the new frontend if available
or the static index.html.


wolfv:
Thanks!

PR #19, opened 2020-07-08
Fix discrepancies from api endpoint change and bulk upload additions

benjamb:
It may be appropriate to remove the `test-cli-client.sh` at this point, however I figured I'll just fix it for now.


wolfv:
wonderful, thanks!

PR #20, opened 2020-07-08
Add .conda format support

nowster:


tom--pollard:
Without [xtarfile](https://pypi.org/project/xtarfile/) being packaged in `conda-forge` I think this implementation makes sense! +1

wolfv:
hi @nowster thanks a lot for the PR!

I am wondering if we should use conda-package-handling here? it's the standalone library that conda uses for handling packages: https://github.com/conda/conda-package-handling

It has the followingn functions to extract (single) files:

https://github.com/conda/conda-package-handling/blob/7c4a4717a6d3a15fd41058d8864df2a750b20322/src/conda_package_handling/api.py#L41-L62

If you only ask for the `info` component, it will skip extracting the package data: 

https://github.com/conda/conda-package-handling/blob/7c4a4717a6d3a15fd41058d8864df2a750b20322/src/conda_package_handling/conda_fmt.py#L52

What do you think? The benefit would be that we use the same code as conda internally...
And one other thing is that the `.conda` file spec doesn't even specifiy zstd as the only compression, but rather it supports any compression algo that comes with libarchive, they just happen to pick zstd for the moment.

nowster:
Maybe not adding a dependency on core conda code when providing alternatives to conda (mamba/boa), perhaps?

Note that we only have to specify a decoder for compression schemes that the `tarfile` module can't automatically handle natively.

wolfv:
to be fair, afaik the conda-package-handling doesn't depend on conda (just libarchive). But the code you have right now is simple enough that I think it's fine to merge it.

If we change our minds later on, we can change the implementation to conda-package-handling.

PR #21, opened 2020-07-08
Handle name not being set in GitHub

benjamb:
Also detect profile changes and update the stored data as a result.

PR #27, opened 2020-07-14
feat: setup.py for use in conda recipe

mariobuikhuizen:


setup.py
```diff
@@ -0,0 +1,20 @@
...
+
+setuptools.setup(
+    name="quetz",
+    version=version_ns['__version__'],
+    author="The Quetz Development Team",
+    description="The Mamba package server",
+    long_description="The Mamba package server",
+    url="https://github.com/thesnakepit/quetz",
+    packages=setuptools.find_packages(),
+    python_requires='>=3.6',
```
benjamb, line 19: As part of #21 I introduced a dependency on python 3.7 through use of postponed evaluation of annotations.

mariobuikhuizen:
Thanks for catching this! I will fix it.

PR #29, opened 2020-07-20
Add support for loading settings from toml config files

### LINKED ISSUE #25: Parse config from a file
Is there are preferred format (yaml, json, ini, toml)?

While this could perhaps be a separate issue, it would also be nice to be able to have configs parsed as command line options (configargparse would be good for this).

uvicorn could then be run programmatically, with the port/listen address (and other settings) loaded from the config.

benjamb:
Ultimately decided against running via uvicorn programmatically, primarily since we'd lose the ability to autoreload during development, though there are additional downsides (see: https://github.com/encode/uvicorn/issues/168#issuecomment-408838644).

As such using configargparse no longer made sense. Tried out a couple of other libraries before deciding on writing something using just toml and appdirs.

Still supports setting environment variables via `QEUTZ_SECTION_NAME`, i.e. `QUETZ_GITHUB_CLIENT_ID=foo`.

On *nix will load settings from `config.toml` files found in `/etc/quetz/`, `~/.config/quetz/` and additionally from the file set via the `QUETZ_CONFIG_FILE` environment variable (in that order).

Closes #25


mariobuikhuizen:
Awesome! Thanks!

PR #40, opened 2020-07-28
Document installing with python 3.7 or later

nowster:


wolfv:
Hi @nowster, can you change this so that it reads `python=3.7`? I think installing python 3.7 is the best option because only with 3.7 you're guaranteed to get the latest conda-build version :)

nowster:
You get python 3.8. And the existing dependency in `setup.py` matches this, and you get python 3.8 with that, too.

PR #42, opened 2020-07-29
Fix Readme tutorial

adriendelsalle:
Description
--

The current tutorial in Readme is broken

``` bash
$ uvicorn quetz.main:app --reload
...
RuntimeError: Directory 'channels' does not exist
```

``` bash
$ ./test-cli-client.sh
./test-cli-client.sh: 2: ./test-cli-client.sh: curl: not found
```

Proposition
--
Add curl dependency 
Create channels directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Then create an environment:
 
 ```
 mamba create -n quetz -c conda-forge python fastapi authlib httpx=0.12.0 sqlalchemy sqlite \
-python-multipart uvicorn zstandard conda-build appdirs toml
+python-multipart uvicorn zstandard conda-build appdirs toml curl
 
 conda activate quetz
 ```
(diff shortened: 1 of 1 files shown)


benjamb:
The `test-cli-client.sh` script is ultimately obsolete now there is the quetz-client tool, so it might make more sense to simply remove that rather than pull in curl for it.

It also might be worth creating the channel directory during app startup (prior to mounting in `main.py`), rather than an explicit step for the user.

adriendelsalle:
Hi @benjamb !

Thanks for your comment, I modified the proc to:
- install `quetz-client` instead of `curl`
- use `quetz-client` to push packages (requires the API token stored in a env variable for now)

The package repository bootstrap could be done by a single command. Maybe I missed it, I will open a dedicated issue otherwise.
Is it OK for you that way?
Thanks!

wolfv:
looks good to me!

wolfv:
thanks for the contribution

PR #50, opened 2020-08-09
add pagination and search to front- and backend for packages

wolfv:
This adds an initial API for pagination to the backend, and uses it in the frontend.
Also adds a search thing to the frontend and uses the `q` parameter to search in the backend.

This does break the API for `/api/channels/{}/packages` though. So I am wondering if we should activate this behind a query parameter (`&pagination=true`) or if we should give it a more generic term?

With the current API we can't know the total amount of available packages. So we should either return _all_ packages or add that value somehow.

### DIFF
--- a/quetz/dao.py
+++ b/quetz/dao.py
@@ -1,7 +1,7 @@
 # Copyright 2020 QuantStack
 # Distributed under the terms of the Modified BSD License.
 
-from sqlalchemy.orm import Session, joinedload, aliased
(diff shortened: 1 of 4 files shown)


dhirschfeld:
I guess with pagination it might be nice to be able to choose the number of results to return - e.g. `num_results

wolfv:
Yep, that's possible with offset and limit parameters 🙂

dhirschfeld:
Ah yep, sorry - should actually read the source before commenting! 😬

PR #55, opened 2020-08-17
Updates to config handling

benjamb:
* Fix unnecessary use of f-strings
* Update config.py with a Config class
* Add support for optional sections
* Assume an entry is not required if it specifies a default
* Fix bool cast
* Explicitly check defaults for None

### DIFF
--- a/init_db.py
+++ b/init_db.py
@@ -3,16 +3,16 @@
 
 import random
 import uuid
-from quetz.config import load_configs
+from quetz.config import Config
 from quetz.database import init_db, get_session
 from quetz.db_models import (User, Identity, Profile, Channel, ChannelMember, Package,
                              PackageMember, ApiKey)
 
 
 def init_test_db():
-    load_configs()
-    init_db()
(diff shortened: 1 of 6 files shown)


wolfv:
this looks great to me.
do you want to fix the whitespace-before-colon thing in the docstrings? It seems like it's different in several places right now.
I don't see a reasoning for having a whitespace before the colon (as stated in the numpydoc). I would hope that parsers of docstrings don't really care about that and I think a whitespace before the colon seems weird.

I am happy to merge as-is though, and we can fix it in a subsequent PR as well.

benjamb:
@wolfv I would hope so too, however I have updated the series to be consistent with what the numpydoc stated.

adriendelsalle:
I never tested if a missing whitespace causes a different rendering, just tried to follow numpydoc guidelines.
I also prefer docstrings without this whitespace before colon.. we could remove it everywhere another time if it makes no difference.
Thanks @benjamb @wolfv

PR #57, opened 2020-08-25
Sphinx doc

adriendelsalle:
Build the documentation using Sphinx and the `pydata` theme.

This first implementation is broadly inspired by [`Jupyter`](https://jupyter.readthedocs.io/en/latest/), [`pandas`](https://pandas.pydata.org/docs/index.html) and [`bokeh`](https://docs.bokeh.org/en/latest/index.html) docs.

Discussion
---
- At this time, a small part of the code is covered by docstrings
- *autodoc* has to be improved to display nicely classes/methods/attributes/etc. in `API reference`
- Configuration options to be detailed in `User Guide`


quetz/cli.py
```diff
@@ -278,11 +278,11 @@ def start(path: str,
 
     Parameters
     ----------
-    path : str
+    path: str
```
benjamb, line 281: I thought we'd agreed to keep the spaces before the `:`?

adriendelsalle: Hi @benjamb 

You're right! I told you it is my own habit to write docstrings without whitespace-before-colon..

This only makes the built html having/not having thoses whitespaces, so we can just remove them everywhere if you prefer.

adriendelsalle: I fixed the cli file. Thanks for having noticed that!
Let me know if you prefer to change the docstring style everywhere.

benjamb: I'm not that bothered myself tbh, though I agree with @wolfv that consistency is what's important.

adriendelsalle: I also agree!

wolfv:
we've added docs now (using sphinx: #127), and they are available at https://quetz.readthedocs.io/ )

adriendelsalle:
Cool!
Sorry this draft PR got stale, I didn't took time to request a review.

wolfv:
no problem at all!

PR #58, opened 2020-08-25
Fix auth regression

benjamb:
Fixes github auth regression introduced in #55 (missing db_url parameter to get_session()).

Thanks to @tom--pollard for spotting.


tom--pollard:
+1 , thanks @benjamb

tom--pollard:
Are you ok with this @wolfv ? :smiley:

PR #59, opened 2020-08-25
Storage and indexing

nowster:
This patch set abstracts storage methods to allow for different storage backends:

* Local filesystem
* S3

It also provides indexing (index.html generation and JSON files) on these backends, and removes the dependency on `conda index`.


wolfv:
great!

Issue #44, opened 2020-07-30
Bootstrap install


adriendelsalle:
Hi!

It could be convenient to create a simple `quetz` CLI (using [Typer](https://typer.tiangolo.com/) ?) to deploy a new Quetz server.

Is there any plan of containeurized Quetz with `docker`, `docker-compose`, `helm`, etc.? 
Is it related to leveraging JHub machinery for this project?

Thank you for this project!

adriendelsalle:
Hi!

@wolfv @SylvainCorlay, how do you prefer to handle docker recipes for this repo, but also for all repos in `TheSnakePit` ?
- a dedicated `docker-stacks` repo gathering all the recipes (the [Jupyter](https://github.com/jupyter/docker-stacks) way)
- each repo carries its _dockerfiles_/_docker_/_whatever_ directory for its recipe(s) (the [JupyterHub](https://github.com/jupyterhub/jupyterhub) way)

Thanks!

PR #60, opened 2020-08-25
Docker image

adriendelsalle:
Description
---
Create a `docker` image to use `Quetz`. 
Containerization is a different way to use an application, see #44 .

I also created a repo for `Quetz` under the `TheSnakePit` organization on DockerHub.
- If it's OK, I will transfer the ownership to QuantStack members
- If not, I will delete it

TODOs
---

- [ ] Update the package requirement in `Dockerfile` to use the CLI
- [ ] Transfer the ownership of the [TheSnakePit DockerHub org](https://hub.docker.com/r/thesnakepit/quetz) to @wolfv or other `QuantStack` members
- [ ] Add GitHub action(s) for building the image and publishing it. Maybe it implies a pipeline to build the conda package and publishing it on conda-forge if the Dockerfile relies on a conda package installation
Discussion
---
Base `Quetz` and `Boa` images on `Mamba` image to be released.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,0 +1,11 @@
+FROM condaforge/miniforge3:latest
+
+ARG QUETZ_VERSION=0.0.4
+
+RUN conda install -q -y quetz==${QUETZ_VERSION} && \
+    conda clean --all -f -y
+ENV LANG=en_US.UTF-8
+
+USER nobody
+# TODO: use the CLI when available in the next release to set-up Quetz
+CMD ["/bin/bash"]


dockerfiles/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM continuumio/miniconda3:latest
```
benjamb, line 1: Should this perhaps be condaforge/miniforge3?

adriendelsalle: Hi @benjamb
I don't understand the comment. Could you please provide me more details about what you suggest to do?
Thanks!

benjamb: Apologies. I was wondering if we should use `condaforge/miniforge3:latest` as the base image? This is a point for discussion though.

adriendelsalle:
Good question, I never used miniforge.

It seems that the miniforge image is based on the Ubuntu bionic image while miniconda3 is based on a Debian one. I guess Debian is lighter.
Does miniforge add anything compared to miniconda?

We should keep a default one for the moment then switch to a mamba or umamba as soon as available don't you think?

dockerfiles/Dockerfile
adriendelsalle: Good question, I never used miniforge.

It seems that the miniforge image is based on the Ubuntu bionic image while miniconda3 is based on a Debian one. I guess Debian is lighter.
Does miniforge add anything compared to miniconda?

We should keep a default one for the moment then switch to a mamba or umamba as soon as available don't you think?

tom--pollard: One benefit is that miniforge installers have conda-forge as the default channel

adriendelsalle: Thanks! I saw that too.

I will dig a bit what's the base layers sizes to keep it as minimale as possible.
That's much more important for pulling time and sparing resources than adding the -c conda-forge or a specific .condarc in the Dockerfile.

benjamb: The miniforge3 image just so happens to be ⅔ the size of the miniconda3 image.

A µmamba image would be very cool though.

adriendelsalle: Thanks you were faster than me ;). Uncompressed, the miniforge3 image is still 2/3 of the miniconda3.
I was surprised the bionic image to be also lighter than the Debian one.

Let's switch to the miniforge3 ! Thanks @benjamb @tom--pollard

wolfv:
looks good. I wonder if we can also add e.g. a github action to somehow automate this process?
One issue is that with this model, we need to wait for the release to be available through conda before we can push the new docker image. 
Maybe we should build the docker image from source, and then run a github action whenever a new tag is available and push it to the registry?

adriendelsalle:
We should create a GitHub action yes!
I'm not familiar with this (more used to GitLab) and I haven't the permissions for that on the repo.
We should do that when after having modified the entry point to use the CLI recently introduced

Jupyter images are built exactly the same way using conda packages so we should understand how they are working in there pipeline to wait for the conda package to be available.
Building from sources is also an option. I prefer using the artefact to force consistency.

beenje:
This PR has been in draft for quite some time. I needed a docker image, so I created a new Dockerfile and made a new PR: https://github.com/mamba-org/quetz/pull/393

My personal opinion is that building from source is not an issue. When you can have the Dockerfile in the source repository, that's actually what you should do.
If you want to use a released version (like the conda package), you should have the Dockerfile in its own repository. But I think it's a bit more work to have a separate repository when you manage the official one.

wolfv:
we have a nice docker file for now from @beenje :)

PR #61, opened 2020-08-26
Add Google OAuth 2.0 OpenID Connect support

tom--pollard:
**!!A few things to tick off before merging!!**

- [x] It would be great if a member of QuantStack/Maintainers could provide an id/secret for the `dev_config.toml`, I believe the current credentials for GitHub are linked to a @mariobuikhuizen endpoint. If so I will add to the PR before merge.

- [x] Google doesn't guarantee a `login` (i.e username/nickname) field like GitHub provides, as such I've gone with the users email for the same DB field (see the [OIDC Standard](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) & what Google will [provide](https://accounts.google.com/.well-known/openid-configuration)). With this is mind, do we want to align `login/username == email` internally in quetz for all $providers as we already ask for it in the GitHub Scope? Outside of OAuth it's probably more reliable to assume we'd receive email addresses over usernames too imo.

- [x] There is no support for having multiple auth providers for the same user Quetz user. I believe it would be nice for a single Quetz user account to auth with N providers and we already store a `provider` value which could be expanded on. Right now I don't think it's much of an issue and could be worked on in a single module, but still worth considering. This module could also be the entry point for a public API providing a mechanism for adding further auth providers/systems to Quetz.

- [x] The Google section of config is set to be required (although entries/fields can be left empty) as with GitHub, with recent changes to [Config](https://github.com/TheSnakePit/quetz/commit/9b35fc487504950e6cb5393cb67420c031631275) we could allow the section to be optional.

**The points above have been covered in the comments. For now Google Auth will be optional config, and further work will be done to integrate multi auth per user going forward**

This could be considered part of https://github.com/TheSnakePit/quetz/issues/41

### DIFF
--- a/quetz_frontend/src/components/Header.vue
+++ b/quetz_frontend/src/components/Header.vue
@@ -14,8 +14,11 @@
         <img class="avatar-img" :src="avatar_url"  />
       </template>
       <template v-else>
-        <cv-button v-on:click="signin">
-          Sign In
+        <cv-button v-on:click="signinGithub">
+          Sign In Via Github
+        </cv-button>
+        <cv-button v-on:click="signinGoogle">
+          Sign In Via Google
         </cv-button>
       <cv-header-global-action aria-label="User avatar" aria-controls="user-panel">
         <!-- <img :src="avatar_url" v-if="avatar_url" /> -->
@@ -58,9 +61,13 @@
       this.me();
     },
     methods: {
-      signin() {
+      signinGithub() {
         window.location.href = "/auth/github/login";
-        console.log("Signing in");
+        console.log("Signing in via github");
+      },
+      signinGoogle() {
+        window.location.href = "/auth/google/login";
+        console.log("Signing in via google");
       },
       logout() {
         window.location.href = "/auth/logout";
--- a/README.md
+++ b/README.md
@@ -129,6 +129,18 @@ url = "https://..."
 
 Be sure to set the url field if not using AWS.
 
+## Google OAuth 2.0 OpenID Connect
+
+To enable auth via Google, you will need to register an application at: https://console.developers.google.com/apis/credentials
+
+Then add the client secret & ID to a `google` section of your `config.toml`:
+
+```
+[google]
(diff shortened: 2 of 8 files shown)


quetz_frontend/src/components/Header.vue
```diff
@@ -14,8 +14,11 @@
         <img class="avatar-img" :src="avatar_url"  />
       </template>
       <template v-else>
-        <cv-button v-on:click="signin">
-          Sign In
+        <cv-button v-on:click="signinGithub">
```
wolfv, line 17: shoudl we have a config setting to enable / disable providers?

wolfv:
Should we require the email instead of username from all OAuth providers? I think that would give us an easier way of merging accounts by email.

On the other hand, merging could be done by first logging in with one, then with the other.

From my perspective I am happy to merge this. We can iron out things here and there later on (e.g. disabling, enabling providers in config, and supporting users with multiple auth mechanisms).

wolfv:
@tom--pollard I don't want to bother you but do you think we can tick the boxes soon. I wanted to make a quetz release and it would be awesome to have these changes in.

wolfv:
I tried your branch just now. I think marking the config for google optional would be appreciated because otherwise existing configs will break.

tom--pollard:
Hi @wolfv sorry yesterday was a public holiday in the UK, I will look to rectify your comment today :smile:

wolfv:
no worries! I hope you had an enjoyable holiday :)

quetz_frontend/src/components/Header.vue
tom--pollard: I've just force pushed to make the google auth section optional, as well as only loading the router/module if configured.

I must admit I know very little Vue. Does it have a method to know if a given route exists (I see it does for vue routers, I don't know how that maps to fastapi/starlette though), maybe that could make these buttons dynamic with `v-if`?

I did think of doing a fetch on the route to see if it's 404's, but that sounds messy

wolfv: indeed an interesting question! Ideally this would be pre-rendered on the server side ... 
I'll merge this for now and think about this tomorrow.

PR #74, opened 2020-09-01
README improvement & examples simplification

alanderex:
The example link for dummyuser renders (http://localhost:8000/api/dummylogin/%5B) and returns Internal Server Error. Should rather be one of the test users like alice.


wolfv:
Thanks for your contribution!

PR #78, opened 2020-09-02
cli help messaging improvements

wolfv:


quetz/cli.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2020 QuantStack
 # Distributed under the terms of the Modified BSD License.
+# flake8: noqa
```
benjamb, line 3: I suspect it would be better to tell flake8 you don't care about line length, rather than disable all checks for the entire file.

I set it to 88 to match the default that black uses. The long strings could however just as easily be split over multiple lines.

wolfv: yeah unfortunately black didn't split the strings in this file :/

PR #79, opened 2020-09-03
cli: General improvements

benjamb:
Originally set out to improve the help messages within cli.py, ended up fixing a few other things as well.

Commits:
cli: Fix line lengths
cli: Fix 'deployement' typo
cli: Fix order of args to create()
cli: Use enum for logging levels
cli: Simplify check for non-config files
cli: Ensure errors are written to stderr
cli: Simplify create() logic a little

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -1,6 +1,5 @@
 # Copyright 2020 QuantStack
 # Distributed under the terms of the Modified BSD License.
-# flake8: noqa
 
 import typer
 import os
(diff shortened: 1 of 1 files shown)


wolfv:
awesome. 
One thing that I also liek to fix is run vs start. I think that's confusing

benjamb:
I've seen similar sub-commands before (e.g. Docker) which does add some level of convenience. I don't have particularly strong opinions on it, though the duplication irks me ever so slightly.

PR #83, opened 2020-09-06
Add instructions for creating a channel and generating an API key

ericdill:
Hi @wolfv --

This is my attempt at documenting the steps I needed to take to generate an API key for my user. Thoughts?

E


wolfv:
Thanks a lot @ericdill ! I think this is great.
I'll try to add this functionality to the quetz-cli this week so that we have a more streamlined approach of dealing with API keys.

PR #85, opened 2020-09-06
Update s3 backend to allow for IAM

ericdill:
Relatively small change that allows for using IAM roles instead of access keys to make use of the s3 backend


wolfv:
Thanks!

PR #86, opened 2020-09-06
Init testing using pytest

adriendelsalle:
@wolfv , to continue the discussion about what is the best way to write tests.

Here is a very simple demo of testing `User` class in `db_models` using `pytest`.

I re-used code from [JHub tests](https://github.com/jupyterhub/jupyterhub/tree/master/jupyterhub/tests) and I'm clearly wondering how to re-use more code from this project.

Also in this PR:
- simplify the database creation
- add find method for User
- refactor cli according to changes

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -12,9 +12,10 @@ from typing import Dict, NoReturn
 
 import typer
 import uvicorn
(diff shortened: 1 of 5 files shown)


wolfv:
Awesome, I'll look at this tomorrow!

Issue #51, opened 2020-08-11
Think about cache proxying packages


wolfv:
One interesting idea that came up was to configure quetz in a way that it proxies the (requested) packages for e.g. Anaconda.org

This could be easy to implement: 

- We just need to serve the repodata from the original location
- Download & forward the packages upon request

adriendelsalle:
Great!

Definitely a *must* have for private compagnies using conda/mamba behind a corporate proxy.

You only have then to configure your `.condarc` to use this server instead of public hosted ones, after having whitelisted the proxied repo urls for the Quetz server host.

wolfv:
We have this now :)

PR #98, opened 2020-09-10
Quetz proxy

btel:


wolfv:
great! also your way of doing tests with dependency injection is much better than what I was trying to do so I am going to base my tests on yours :)

adriendelsalle:
#51

Issue #81, opened 2020-09-03
Figure out how to handle private channels


wolfv:
I think while we have private channels, we need to implement how we handle them properly. I think right now every user can see the contents of private channels, for example.

adriendelsalle:
Related to #54 ?

wolfv:
yes but this was also about the `/api/channels` for example. And `/api/channels/channelX/packages`.
I think they should return different things depending on the permission level.

I need to finish this one tricky join today so that this hides private channels that User is not a member of.

adriendelsalle:
Nice!
I agree, this what I called a user *view*. It apply to the repodata.json but also on any other requests to the api.

wolfv:
@adriendelsalle in this PR I have added search, that only returns packages that the logged in user has access to: https://github.com/TheSnakePit/quetz/pull/82

I want to add tests next week so that we can check that this behaves as intended.

adriendelsalle:
#105

Issue #54, opened 2020-08-12
Fine-grained permissions for pulling?


adriendelsalle:
Hi!

I have a doubt concerning the meaning of fine-grained permissions on channels and packages.

While it's pretty clear that pushing is covered to allow flexibility for developers/recipe maintainers/etc., are we also talking about per-user or per-group pulling permissions?

I don't know much about how private channels and user token are currently handled by conda.

It is a major feature for companies who are tied with strict security rules such as access control and segregation, or just simply want to log activity for security purposes (track abnormal pulls from a single user, etc.). 

Thanks!

PR #105, opened 2020-09-14
feat: prevent downloading of private channels

mariobuikhuizen:
Use the same path prefix as conda for specifying a token/api-key


wolfv:
this is great! really cool with the middleware.

adriendelsalle:
#81 #54

PR #108, opened 2020-09-17
fix mirror backfill from channels without channeldata

btel:


quetz/main.py
```diff
@@ -344,13 +345,18 @@ def post_channel(
...
+        except (RemoteFileNotFound, json.JSONDecodeError):
+            subdirs = None
         except RemoteServerError:
             raise HTTPException(
                 status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
                 detail=f"Remote channel {host} unavailable",
             )
+        # if no channel data use popular architectures
+        if subdirs is None:
+            subdirs = ["linux-64", "win-64", "win-32", "osx-64"]
```
wolfv, line 358: I wonder if we can import the list of known platforms from conda?

wolfv (changes requested):
can you squash the pre-commit run? 
The change with importing known platforms would be cool 

https://github.com/conda/conda/blob/61f1836824dc68a518227c4e829ca664d2dd6307/conda/base/constants.py#L63

btel:
sure I will squash the pre-commit runs.

I will check also whether we can get the arch from conda, but this will add dependency on conda?

wolfv:
right i think we removed the depenedency on conda ... you can just copy-paste in that case.

btel:
ok, thanks, for reference here is the list of platforms from conda:

https://github.com/conda/conda/blob/a78a2387f26a188991d771967fc33aa1fb5bb810/conda/base/constants.py#L63

btel:
@wolfv good to merge

PR #113, opened 2020-09-24
implement synchronisation of mirror channels

btel:
This PR implement a simple scheme for synchronization of mirrors.

To synchronize for each package it checks the `time_changed` attribute in the repodata, if the timestamp is later than the timestamp of the last synchronised package (from previous runs), it updates the file, otherwise it skips. If the attribute is not present (for example for anaconda server, it  forces the update).

In addition, this PR adds a new endpoint `PUT /channels/{channel_name}` (without data), which will trigger a synchronisation run (as a background task). Note that synchronisation is not triggered automatically.


wolfv:
For anaconda servers (or others that do not implement the time_changed attribute) we could also compare the tuple `(full_package_name, SHA256 sum)`. For example `(xtensor-0.21.5-h123323.tar.bz2, 123123)`. Is this also implemented in this PR?

btel:
yes, this could work. However, we would have to compare the tuple against quetz database for each package. Perhaps we could do this in a separate PR?

wolfv:
we could also compare it against the latest repodata from quetz vs. anaconda

btel:
that's right. should we try it in the next PR?

wolfv:
yep!

PR #114, opened 2020-09-30
Implement sync for non quetz servers

btel:
as discussed in PR #113 for non-quetz server that do not implement `time_modified` attribute we can synchronize versions by comparing the package/filename and the SHA256 number of the local and remote files.

 This PR implements this feature for syncing mirror channels.


btel:
ready to merge @wolfv

PR #117, opened 2020-10-01
fix postgres compatibility issues

btel:
fix schemas and tests to make quetz and the tests run with postgres db

running with postgres  fixes the problem of segfaults when adding multiple versions of the same package (issue #118)

also add postgres-based tests to the build matrix on CI


btel:
@wolfv good to merge

wolfv:
thanks, this is great!

PR #119, opened 2020-10-08
Onboarding create api keys

btel:
a few fixes to streamline onboarding new users:

* allow create api keys without special privilages, this key is valid for example to post (create) new channels
* return the new key after it is created with the POST /api/api-keys method
* add a button in basic fronted to retrieve and display the key in a modal window
* also some more tests and fixes 

Notes:
* POST /api/api-keys plays two roles: creating new keys and assigning permissions, the permissions are not specific to the key but for all keys of the user
* there is no role of superadmin who by default has read/write access to all channels/packages/users

### DIFF
--- a/quetz/authorization.py
+++ b/quetz/authorization.py
@@ -137,11 +137,14 @@ class Rules:
                     role.package,
                     required_package_role,
                 )
-            else:
+            elif role.channel:
                 required_channel_roles = (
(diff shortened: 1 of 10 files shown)


btel:
@wolfv this is mergeable, but it might be good idea to talk through the changes and plans for the implementation of permission system

PR #120, opened 2020-10-16
Remove roles from api endpoints

btel:
these roles are redundant with the `channels/{}/members` and `packages/{}/members` endpoints.  we plan to add a proper support for scopes in the future  `"api:read", "api:write"`, etc.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,12 @@ client_secret = "EXAMPLESECRETmD-7UXVCMZV3C7b-kZ9yf70"
 
 ## PostgreSQL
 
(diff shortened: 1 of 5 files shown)


btel:
good to merge @wolfv

wolfv:
hm! I think one idea was that we would like to have fine-grained permissions:

API Key per channel / per package (not necessarily tied to a user)
as well as probably API key per user ... 

If I understand correctly, with this change we have only API key per user, right?

btel:
right, that might be interesting

so we could keep the roles in the api keys, but also add a user role so that we can get a wildcard key per user

PR #139, opened 2020-11-05
add conda-suggest plugin

madhur-tandon:
This is without `tests` for now.

To use:
1) Create a new channel.
2) Upload some packages.
3) Hit the endpoint `/api/channels/{channel_name}/conda-suggest`

Maybe we should do creation of the `channel_suggest_map` and the `downloadable map file` as a background task?

### DIFF
--- a/plugins/quetz_conda_suggest/README.md
+++ b/plugins/quetz_conda_suggest/README.md
@@ -1,0 +1,12 @@
+# quetz_conda_suggest plugin
+
+This is a plugin to use with the [quetz](https://github.com/mamba-org/quetz) package server.
+
+
+## Installing
+
+To install use:
+
+```
+pip install .
+```
--- a/plugins/quetz_conda_suggest/quetz_conda_suggest/api.py
+++ b/plugins/quetz_conda_suggest/quetz_conda_suggest/api.py
@@ -1,45 +1,29 @@
-import json
+import os
 
 from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi.responses import FileResponse
 from sqlalchemy.orm.session import Session
 
-from quetz.db_models import PackageVersion
 from quetz.deps import get_db
 
 router = APIRouter()
 
 
-@router.get("/api/channels/{channel_name}/conda-suggest")
-def get_conda_suggest(channel_name, db: Session = Depends(get_db)):
-    all_packages = (
-        db.query(PackageVersion)
-        .filter(PackageVersion.channel_name == channel_name)
-        .all()
+@router.get("/api/channels/{channel_name}/{subdir}/conda-suggest")
+def get_conda_suggest(channel_name, subdir, db: Session = Depends(get_db)):
+    map_filename = "{0}.{1}.map".format(channel_name, subdir)
+    map_filepath = os.path.join(
+        os.getcwd(), "channels", channel_name, subdir, map_filename
     )
 
-    channel_suggest_map = {}
-    error = False
-    for each_package in all_packages:
-        if not each_package.files:
-            error = True
-        else:
-            files_data = json.loads(each_package.files.data)
-            for (k, v) in files_data.items():
-                channel_suggest_map[k] = v
-
-    if error:
+    if os.path.isfile(map_filepath):
+        return FileResponse(
+            map_filepath,
+            media_type="application/octet-stream",
+            filename=map_filename,
+        )
+    else:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
-            detail="files for all packages not found",
+            detail=f"conda-suggest map file for {channel_name}.{subdir} not found",
         )
-
-    with open("{0}.map".format(channel_name), "w") as f:
-        for (k, v) in sorted(channel_suggest_map.items()):
-            f.write("{0}:{1}\n".format(k, v))
-
-    return FileResponse(
-        "{0}.map".format(channel_name),
-        media_type="application/octet-stream",
-        filename="{0}.map".format(channel_name),
-    )
--- a/plugins/quetz_conda_suggest/quetz_conda_suggest/api.py
+++ b/plugins/quetz_conda_suggest/quetz_conda_suggest/api.py
@@ -1,0 +1,45 @@
+import json
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from fastapi.responses import FileResponse
+from sqlalchemy.orm.session import Session
+
+from quetz.db_models import PackageVersion
+from quetz.deps import get_db
+
+router = APIRouter()
+
+
+@router.get("/api/channels/{channel_name}/conda-suggest")
+def get_conda_suggest(channel_name, db: Session = Depends(get_db)):
+    all_packages = (
+        db.query(PackageVersion)
+        .filter(PackageVersion.channel_name == channel_name)
+        .all()
+    )
+
+    channel_suggest_map = {}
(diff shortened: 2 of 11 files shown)


plugins/quetz_conda_suggest/quetz_conda_suggest/main.py
```diff
@@ -0,0 +1,34 @@
...
+    with get_db() as db:
+        if not version.files:
+            metadata = db_models.CondaSuggestMetadata(
+                version_id=version.id, data=json.dumps(suggest_map)
+            )
+            db.add(metadata)
+        else:
+            metadata = db.query(db_models.CondaSuggestMetadata).get(version.id)
+            metadata.data = json.dumps(suggest_map)
+        db.commit()
```
wolfv, line 34: Looks all really good!

Can you write an updated conda-suggest _file_ here? And then, in the get_conda_suggest function above you can just return that file (or it can be handled like a static file).
We do that for repodata.json.

madhur-tandon: This means that we essentially need to maintain a global state for the `channel_suggest_map` in `api.py`. Are you suggesting to maintain that, we create a file (serialization of `channel_suggest_map`) that we update every time a new package is uploaded?

I have one question though --  simple appending it is easy, but we need to maintain the constraint that the file should be sorted always (first by `keys`, then by `values`). So the update could be tricky.

Though, I do agree that incremental update is better than generating a new file always (by scanning all packages) as it's currently done.

madhur-tandon: Also, in that case, would we require `db_models.py` at all? I don't see it's use if we serialize the global state of `channel_suggest_map` in a file.

wolfv: yeah because we could use the database everytime we regenerate the file. but it's true, we wouldn't strictly need it.

wolfv:
there are some additional challenges, btw., if we want to be fully correct

- we need to use only packages of some arch to generate the conda-suggest-map
- we only want the metadata from the latest version in the conda-suggest-map
- Windows looks slightly different
- noarch packages need to parse entrypoints

plugins/quetz_conda_suggest/tests/test_main.py
```diff
@@ -0,0 +1,6 @@
+def test_conda_suggest_endpoint(client):
+    pass
```
btel, line 2: don't forget to add some tests

quetz/condainfo.py
```diff
@@ -106,6 +107,9 @@ def _load_jsons(self, tar):
         self.info = json.load(tar.extractfile("info/index.json"))
         self.about = json.load(tar.extractfile("info/about.json"))
         self.paths = json.load(tar.extractfile("info/paths.json"))
+        with tar.extractfile("info/files") as fp:
```
btel, line 110: you may also need to extract files from .conda format packages

btel (commented):
LGTM, great work!

plugins/quetz_conda_suggest/quetz_conda_suggest/main.py
madhur-tandon: Done!

plugins/quetz_conda_suggest/tests/test_main.py
madhur-tandon: Yes, gotta do it today 👍🏼

quetz/condainfo.py
madhur-tandon: Yes, this works. Just checked.
This is because the `_load_jsons` function is called after both the formats have been handled 👍🏼

plugins/quetz_conda_suggest/tests/test_quetz_conda_suggest.py
```diff
@@ -0,0 +1,77 @@
...
+    with mock.patch("quetz_conda_suggest.main.get_db", get_db):
+        main.post_add_package_version(package_version, condainfo)
+
+    meta = db.query(db_models.CondaSuggestMetadata).all()
+
+    assert len(meta) == 1
+    assert meta[0].data == '{"test-bin": "test-package"}'
+
+
+def test_conda_suggest_endpoint_with_upload(
```
madhur-tandon, line 62: @btel Could use some help in this test. This is where I get `409: Duplicate`.
And `404` for the `conda-suggest` endpoint since map file doesn't exist there for some reason.

plugins/quetz_conda_suggest/tests/test_quetz_conda_suggest.py
```diff
@@ -0,0 +1,67 @@
...
+
+    assert len(meta) == 1
+    assert meta[0].data == '{"test-bin": "test-package"}'
+
+def test_conda_suggest_endpoint_with_upload(client, channel, package, subdir, config, session_maker):
+    response = client.get("/api/dummylogin/madhurt")
+    
+    filename = "test-package-0.1-0.tar.bz2"
+    url = f'/api/channels/{channel.name}/files/'
+    files = [('files', open(filename, 'rb'))]
```
btel, line 69:
```suggestion
    files = [('file', (filename, open(filename, 'rb')))]
```

btel, line 69:
```suggestion
    files = {'file': (filename, open(filename, 'rb')))]
```

plugins/quetz_conda_suggest/quetz_conda_suggest/api.py
```diff
@@ -0,0 +1,29 @@
...
+    map_filename = "{0}.{1}.map".format(channel_name, subdir)
+    map_filepath = os.path.join(
+        os.getcwd(), "channels", channel_name, subdir, map_filename
+    )
+
+    if os.path.isfile(map_filepath):
+        return FileResponse(
+            map_filepath,
+            media_type="application/octet-stream",
+            filename=map_filename,
```
btel, line 23: what if map_filepath does not exist, the handler will return None

madhur-tandon: Then we should get `404` since we check if it exists or not using `os.path.isfile` and we do get `200` as status code. I think we are just unable to get the output from the `response` object.

btel: you're right

plugins/quetz_conda_suggest/README.md
```diff
@@ -0,0 +1,12 @@
+# quetz_conda_suggest plugin
+
+This is a plugin to use with the [quetz](https://github.com/mamba-org/quetz) package server.
```
wolfv, line 3: can you add a line to say that this produces files that can be used with conda-suggest and a link to conda-suggest?

wolfv:
can you squash the linting commits a bit? :)

If you want you can install pre-commit and then the pre-commit hooks and linting will be done everytime you do `git commit`

```
mamba install pre-commit
pre-commit install-hooks
```

plugins/quetz_conda_suggest/README.md
madhur-tandon: Updating it in a second.

wolfv:
Congrats! Thanks for the 3rd quetz plugin :tada:

madhur-tandon:
💃🏼

PR #147, opened 2020-11-10
add users section to config

btel:
implementation of [QEP1](https://quetz.readthedocs.io/en/latest/qeps/qep-001-user-permissions.html) (part 2)


wolfv:
Thanks, this is awesome :)

PR #151, opened 2020-11-13
implement singleton config

btel:
make Config a singleton class, so that we don't have to re-load the config values from the file each time the config is created

Note that for each use of Config, we  instatiate the class (`config = Config()`), which returns the singleton instance. This allows us to simply reset the instance in the tests.


wolfv:
Looks good!

btel:
good to merge @wolfv

PR #152, opened 2020-11-13
Add mypy checks

btel:


wolfv:
Cool! Could this also be done with pre commit? Would that make sense?

btel:
I tried, but I couldn't get it to work. maybe it's also not stupid to have a separate task, to see what fails

PR #153, opened 2020-11-13
Fix cli init db

adriendelsalle:
Hi!

@btel @wolfv 
Seems the CLI is broken due to a database init using not required entries, thus sometimes attributes do not exist.
When passing one of the possible entries (as done in the current tests), the others are defaulting to empty lists, not causing an error.

BTW, the database init should not be done at `create` step and not `run` one ?
Calling cli command `quetz create ...` currently won't init the database.

My patch for the test fixture is maybe not ideal, I use the `pytest.mark.parametrize` decorator only to pass 1 argument/value.
Let me know if you have better ideas ;)

Thanks for your feedback!

### DIFF
--- a/quetz/tests/test_cli.py
+++ b/quetz/tests/test_cli.py
@@ -13,9 +13,12 @@ def user_group():
 
 @pytest.fixture
 def config_extra(user_group):
-    return f"""[users]
-    {user_group} = ["bartosz"]
-    """
+    if user_group is None:
+        return ""
+    else:
+        return f"""[users]
+                {user_group} = ["bartosz"]
+                """
 
 
 @pytest.mark.parametrize(
@@ -23,6 +26,7 @@ def config_extra(user_group):
     [("admins", "owner"), ("maintainers", "maintainer"), ("members", "member")],
 )
 def test_init_db(db, config, config_dir, user_group, expected_role):
+
     def get_db(_):
         return db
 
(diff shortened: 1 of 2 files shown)


btel:
Hi @adriendelsalle , thanks for the catchy.  Your PR looks good. The  use of parametrize is legitimate,  I often use it to substitute fixtures like you did.  I agree about putting initdb in the create instead of run

btel (approved):

quetz/tests/test_cli.py
```diff
@@ -13,40 +13,46 @@ def user_group():
...
+                """
 
 
-@pytest.mark.parametrize(
-    "user_group,expected_role",
-    [("admins", "owner"), ("maintainers", "maintainer"), ("members", "member")],
-)
-def test_init_db(db, config, config_dir, user_group, expected_role):
+def get_user(db, config_dir):
+
```
btel, line 25: You Need to remove this empty line to satisfy linter
```suggestion
```

btel (approved):

quetz/tests/test_cli.py
adriendelsalle: Thanks! I've just made another commit to move the init_db and remove this empty line :)

btel: Cool, thanks!

wolfv:
thanks @adriendelsalle !

PR #156, opened 2020-11-16
fix listing users without profile

btel:
fixes the bug introduced in #147


wolfv:
will this hide users without privileges?

Issue #160, opened 2020-11-17
API permission errors


hbcarlos:
Logged in without a role I can do:

- [ ] See members from channels
<img width="1439" alt="listUsers" src="https://user-images.githubusercontent.com/26092748/99409258-76c84d80-28f1-11eb-90b0-e7b5e8602fae.png">

- [ ] Add packages to a channel
<img width="1443" alt="addPackage" src="https://user-images.githubusercontent.com/26092748/99409292-80ea4c00-28f1-11eb-8f0e-77d2ad70f35e.png">

When trying to upload a file there is an error message `Internal Server Error` in both endpoints.
I tried to upload a pdf to package `new` channel `two test` and my user doesn't have a role.
<img width="1442" alt="Screenshot 2020-11-17 at 16 24 57" src="https://user-images.githubusercontent.com/26092748/99409325-89428700-28f1-11eb-8ea1-89a287aef213.png">

btel:
fixed by #164, #165 and #167

PR #167, opened 2020-11-18
handle errors in package format in condainfo

btel:
related to #160


wolfv:
should we also error out if a file doesn't end with `.tar.bz2` or `.conda`? Not sure if we're doing this currently.

btel:
i don't think so

PR #168, opened 2020-11-18
update  CI workflow

btel:
CI would not fail if only one of the testing commands fails

also fixing some failing tests

### DIFF
--- a/quetz/tests/test_auth.py
+++ b/quetz/tests/test_auth.py
@@ -45,7 +45,7 @@ class Data:
         self.package2 = Package(name="Package2", channel=self.channel2)
 
         self.channel_member = ChannelMember(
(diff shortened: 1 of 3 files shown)


quetz/tests/test_auth.py
```diff
@@ -498,9 +497,9 @@ def test_use_wildcard_api_key_to_authenticate(data, client):
 
     # using per-channel key
 
-    response = client.get(
-        "/api/channels/privatechannel/members", headers={"X-API-Key": channel_key}
-    )
+    # response = client.get(
```
wolfv, line 500: is this a regression?

btel: oups, forgot to uncomment

PR #179, opened 2020-11-20
Subprocess worker

btel:
* implementation of worker launching tasks in sub-processes
* adds unit tests for workers
* refactors worker class
* adds dependency on starlette-full
* fixes issues with testing postgres, where it might fail silently
* disables unused plugins for testing which avoids db consistency issues with postgres


btel:
@wolfv ready to merge

wolfv:
nice cleanups, too!

PR #182, opened 2020-11-24
Add repodata_zchunk plugin

davidbrochart:
This PR implements partial download of the repodata using [zchunk](https://github.com/zchunk/zchunk). It should dramatically reduce repodata downloads when you already have an older repodata locally (which is the case most of the time). It will also require some work in mamba.
It's very much a work in progress, mainly to get early feedback.

### DIFF
--- a/plugins/quetz_repodata_zchunk/quetz_repodata_zchunk/main.py
+++ b/plugins/quetz_repodata_zchunk/quetz_repodata_zchunk/main.py
@@ -1,0 +1,102 @@
+import bz2
+import hashlib
+import json
+import os
+import subprocess
+import tempfile
+from datetime import datetime, timezone
+
+import quetz
+from quetz.tasks.indexing import _jinjaenv
+
+
+def update_index(pkgstore, updated_files, channel_name, subdir, packages):
+    "update index.html in platform subdir"
+
+    add_files = []
+
+    for fname, data in updated_files:
+        md5 = hashlib.md5()
+        sha = hashlib.sha256()
+        if not isinstance(data, bytes):
+            data_bytes = data.encode("utf-8")
+        else:
+            data_bytes = data
+        md5.update(data_bytes)
+        sha.update(data_bytes)
+
+        add_files.append(
+            {
+                "name": f"{fname}",
+                "size": len(data_bytes),
+                "timestamp": datetime.now(timezone.utc),
+                "md5": md5.hexdigest(),
+                "sha256": sha.hexdigest(),
+            }
+        )
+
+    jinjaenv = _jinjaenv()
+    subdir_template = jinjaenv.get_template("subdir-index.html.j2")
+
+    pkgstore.add_file(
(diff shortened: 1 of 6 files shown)


plugins/quetz_repodata_zchunk/quetz_repodata_zchunk/main.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+import subprocess
+import tempfile
+from datetime import datetime, timezone
+
+import quetz
+from quetz.tasks.indexing import _jinjaenv
+
+
+def update_index(pkgstore, updated_files, channel_name, subdir, packages):
```
btel, line 13: we discussed with @madhur-tandon  to  re-implement the update index function in the main tree using the new pkgstore.list_files function. This would avoid re-doing the index in each plugin. @madhur-tandon did you try this?

plugins/quetz_repodata_zchunk/tests/test_main.py
```diff
@@ -0,0 +1,155 @@
...
+    package_file_name,
+    dao,
+    db,
+):
+    @contextmanager
+    def get_db():
+        yield db
+
+    with mock.patch("quetz_repodata_patching.main.get_db_manager", get_db):
+        indexing.update_indexes(dao, pkgstore, channel_name)
```
btel, line 122:
```suggestion
    indexing.update_indexes(dao, pkgstore, channel_name)
```

I don't think you need the mock here.

btel (commented):
LGTM

plugins/quetz_repodata_zchunk/quetz_repodata_zchunk/main.py
madhur-tandon: It's done with the other way of using a `files` dict where each key is a `subdir` and the corresponding `value` is a list of all the files associated with the `subdir`. It's better than `pkgstore.list_files` because:
1) `pkgstore.list_files` gives us all the filenames and not their contents. So we will first have to filter which files do we really want (according to `subdir`).
2) Then, since it only returns filenames, we would need to extract raw bytes data out of each file to calculate the `sha256` and `md5` sum. But, extraction of raw bytes is different for different types of files (`.json`, `.json.bz2`, `.tar.bz2`, `.map`, etc.). Further, we need not do this for packages (`.tar.bz2`) since their hashes are already there in `repodata.json`. So it gets more complicated.

Thus, a very simple straightforward thing to do is to implement the `post_package_indexing` hook and make a function call to the `add_entry_for_index` function. Just like it's done here: https://github.com/mamba-org/quetz/pull/185/files#diff-bc7de45aaa4b34886df14bb6c346ed1e79c5aaf3bc3655f36fc5de9fd9885297R36

wolfv:
just merged @madhur-tandon PR so the tests would hopefully pass now.

davidbrochart:
Trigger CI

PR #184, opened 2020-11-25
add plugin installation to quetz_cli

wolfv:


quetz/cli.py
```diff
@@ -483,5 +485,30 @@ def list() -> NoReturn:
...
+        assert (abs_path / "setup.py").exists()
+
+        exes = ['micromamba', 'mamba', 'conda']
+        if (abs_path / "requirements.txt").exists():
+            for exe in exes:
+                exe_path = find_executable(exe)
+                if exe_path:
+                    break
+
+        if not exe_path:
```
btel, line 502: what if requirements.txt does not exist?

quetz/cli.py
```diff
@@ -483,5 +485,30 @@ def list() -> NoReturn:
         typer.echo('\n'.join([p for p in deployments]))
 
 
+@app.command()
+def plugin(cmd: str, path: str = typer.Argument(None, help="Path to the plugin folder")) -> NoReturn:
+
```
btel, line 490: it might be also cool to install directly from github:

```
quetz plugin install btel/quetz-plugin
```

quetz/cli.py
```diff
@@ -483,5 +485,30 @@ def list() -> NoReturn:
...
+    if cmd == 'install':
+        abs_path = Path(path).absolute()
+        assert (abs_path / "setup.py").exists()
+
+        exes = ['micromamba', 'mamba', 'conda']
```
btel, line 495: should we also include pip?

btel (commented):
looks great! Just a few comments

quetz/cli.py
wolfv: included pip

quetz/cli.py
wolfv: yeah that would be cool ...

quetz/cli.py
wolfv: handled that case now

PR #185, opened 2020-11-26
current-repodata plugin + update conda-suggest and repodata-patching plugin

madhur-tandon:
1. A new plugin `quetz_current_repodata` without an `api.py` i.e. no endpoint. Thus, it only contains the `hook`
2. The `index.html` is generated after all such `hooks` are called and the `files` parameter is updated (defined in `indexing.py`). This is different from using `pkgstore.list_files` since that would also give us data for multiple `subdirs` + would also list `packages` -- which is all extra information.
3. No tests are written for the new plugin `quetz_current_repodata` since there is no API endpoint.
4. The `quetz_conda_suggest` plugin is updated so as to use `pkgstore.add_file()` to create the `.map` file. It also updates the `index.html` so that `.map` appears in it.

### DIFF
--- a/plugins/quetz_current_repodata/quetz_current_repodata/main.py
+++ b/plugins/quetz_current_repodata/quetz_current_repodata/main.py
@@ -9,7 +9,7 @@ from quetz.utils import add_entry_for_index
 
 @quetz.hookimpl
 def post_package_indexing(
-    pkgstore: "quetz.pkgstores.PackageStore", channel_name, subdirs, files
+    pkgstore: "quetz.pkgstores.PackageStore", channel_name, subdirs, files, packages
 ):
     fname = "current_repodata.json"
     pins = {}
--- a/plugins/quetz_current_repodata/quetz_current_repodata/main.py
+++ b/plugins/quetz_current_repodata/quetz_current_repodata/main.py
@@ -1,0 +1,67 @@
+import bz2
+import hashlib
+import json
+from datetime import datetime, timezone
+
+from conda_build.index import _build_current_repodata
+from fastapi import APIRouter
+
+import quetz
+
+router = APIRouter()
+
+
+@quetz.hookimpl
+def register_router():
+    return router
+
+
+@quetz.hookimpl
+def post_package_indexing(
+    pkgstore: "quetz.pkgstores.PackageStore", channel_name, subdirs, files
+):
+    fname = "current_repodata.json"
+    pins = {}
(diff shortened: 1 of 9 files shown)


plugins/quetz_current_repodata/quetz_current_repodata/main.py
```diff
@@ -0,0 +1,67 @@
...
+import hashlib
+import json
+from datetime import datetime, timezone
+
+from conda_build.index import _build_current_repodata
+from fastapi import APIRouter
+
+import quetz
+
+router = APIRouter()
```
wolfv, line 11: unused?

madhur-tandon: Yeah, I removed it before but then the below `register_router` hook needs it. OR can I remove the hook below as well?

plugins/quetz_current_repodata/quetz_current_repodata/main.py
```diff
@@ -0,0 +1,67 @@
...
+
+        updated_files.append((fname, raw_current_repodata))
+        updated_files.append((fname + ".bz2", compressed_current_repodata))
+
+        update_index(pkgstore, updated_files, channel_name, subdir, files)
+
+
+def update_index(pkgstore, updated_files, channel_name, subdir, files):
+    for fname, data in updated_files:
+        md5 = hashlib.md5()
```
wolfv, line 50: maybe we can factor this md5 & sha256 computation, and adding the file to `files` into a function? I guess we're doing this from a couple of places now.

madhur-tandon: Done!

plugins/quetz_current_repodata/quetz_current_repodata/main.py
wolfv: I think so :)

wolfv:
nice!

plugins/quetz_current_repodata/quetz_current_repodata/main.py
madhur-tandon: Done!

wolfv:
hey, congrats! Looks good to me :+1:

Do you want to squash the commits a little bit? Gonna merge right after.

madhur-tandon:
Do you need README.md + tests for the `current-repodata` plugin?

wolfv:
ideally, yes :)

Issue #84, opened 2020-09-06
Failed package uploads still appear in repodata.json


ericdill:
Took me a while to get the s3 backend working. Resulted in lots of failed attempts at uploading. In spite of the package failing to upload due to s3 permission errors, the package still appears in the repodata.json. Quetz should probably be smart enough to not update the repodata.json if the package fails to upload to s3

ericdill:
Another question about the intended usage of the s3 backend: It looks like the implementation is designed as one bucket per channel?

wolfv:
Thanks for this bug report! Very valuable.

wolfv:
should be fixed now, please report if it's still an issue.

PR #191, opened 2020-11-28
concurrency: rollback transaction if package already exists

btel:


wolfv:
does this fix https://github.com/mamba-org/quetz/issues/84 ?

PR #194, opened 2020-11-30
add redis queue support

madhur-tandon:


madhur-tandon:
@btel This should supposedly work. Ignore the repeated wrapper function for now and all other refinements. But, for now, I get this error: `TypeError: cannot pickle '_thread.lock' object`

quetz/tasks/workers.py
```diff
@@ -140,3 +139,65 @@ async def wait(self):
...
+        self.host = host
+        self.port = port
+        self.db = db
+        self.api_key = api_key
+        self.browser_session = browser_session
+        self.conn = redis.StrictRedis(host=self.host, port=self.port, db=self.db)
+        self.queue = Queue(connection=self.conn)
+
+    @staticmethod
+    def wrapper(self, func, api_key, browser_session, **kwargs):
```
btel, line 155: static methods don't take the self, argument so it should rather be:

```suggestion
    def wrapper(func, api_key, browser_session, **kwargs):
```

btel (commented):
it looks good to me. I found one issue that might explain your problem with the serialization.

quetz/deps.py
```diff
@@ -90,12 +90,17 @@ def get_tasks_worker(
...
+        worker = "redis-queue"
 
     if worker == "thread":
         worker = ThreadingWorker(background_tasks, dao, auth, session, config)
+        print("Threading Worker")
     elif worker == "subprocess":
         worker = SubprocessWorker(auth.API_key, auth.session)
+        print("Subprocess Worker")
+    elif worker == "redis-queue":
+        worker = RQManager('127.0.0.1', 6379, 0, auth.API_key, auth.session)
```
btel, line 102: the ip and port of the rq manager should be taken from config, but you can handle it when you finalize the PR

madhur-tandon: Yeah, this is for later.

quetz/tasks/workers.py
madhur-tandon: Yes, but even removing this doesn't work. See the above comment.

PR #198, opened 2020-12-01
add tests for redis

madhur-tandon:
For some reason, the tests fail. But, do these changes suffice for adding the test? What about the `async` and `await` keywords? Are they applicable here? CC: @btel @wolfv 

Also, they won't run on CI for now. So the above comment was for failing locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,6 @@ jobs:
 
           echo "testing server"
 
-          rq worker &
-
           pytest -v ./quetz/tests/
 
           echo "install and test plugins"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,15 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+      redis:
+        image: redis
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 6379:6379
     steps:
       - uses: actions/checkout@v2
 
@@ -63,6 +72,9 @@ jobs:
           export QUETZ_IS_TEST=1
 
           echo "testing server"
+
+          rq worker &
+
           pytest -v ./quetz/tests/
 
           echo "install and test plugins"
--- a/quetz/deps.py
+++ b/quetz/deps.py
@@ -95,15 +95,9 @@ def get_tasks_worker(
     if worker == "thread":
         worker = ThreadingWorker(background_tasks, dao, auth, session, config)
     elif worker == "subprocess":
-        worker = SubprocessWorker(auth.API_key, auth.session)
+        worker = SubprocessWorker(auth.API_key, auth.session, config)
     elif worker == "redis-queue":
-        worker = RQManager(
-            config.redis_ip,
-            config.redis_port,
-            config.redis_db,
-            auth.API_key,
-            auth.session,
-        )
+        worker = RQManager(auth.API_key, auth.session, config)
     else:
         raise ValueError("wrong configuration in worker.type")
 
--- a/quetz/deps.py
+++ b/quetz/deps.py
@@ -92,12 +92,23 @@ def get_tasks_worker(
     else:
         worker = "thread"
 
+    if config.configured_section("redis"):
+        redis_ip = config.redis_ip
+        redis_port = config.redis_port
+        redis_db = config.redis_db
+    else:
+        redis_ip = "127.0.0.1"
+        redis_port = 6379
+        redis_db = 0
+
     if worker == "thread":
         worker = ThreadingWorker(background_tasks, dao, auth, session, config)
     elif worker == "subprocess":
         worker = SubprocessWorker(auth.API_key, auth.session, config)
     elif worker == "redis-queue":
-        worker = RQManager(auth.API_key, auth.session, config)
+        worker = RQManager(
+            redis_ip, redis_port, redis_db, auth.API_key, auth.session, config
+        )
     else:
         raise ValueError("wrong configuration in worker.type")
 
--- a/quetz/tasks/workers.py
+++ b/quetz/tasks/workers.py
@@ -157,7 +157,8 @@ class RQManager(AbstractWorker):
             wrapper, func, self.api_key, self.browser_session, *args, **kwargs
         )
 
-    def wait(self):
+    async def wait(self):
         while not self.job.is_finished:
             time.sleep(1)
-        return self.job.result
+        if self.job.result:
+            return await self.job.result
--- a/quetz/tasks/workers.py
+++ b/quetz/tasks/workers.py
@@ -29,7 +29,7 @@ def prepare_arguments(func: Callable, **resources):
     return kwargs
 
 
-def wrapper(func, api_key, browser_session, **kwargs):
+def wrapper(func, api_key, browser_session, config, **kwargs):
 
     # database connections etc. are not serializable
     # so we need to recreate them in the proces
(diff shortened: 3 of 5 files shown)


madhur-tandon:
Update: The tests pass locally, only the CI remains to be configured.

quetz/deps.py
```diff
@@ -95,15 +95,9 @@ def get_tasks_worker(
...
+        worker = SubprocessWorker(auth.API_key, auth.session, config)
     elif worker == "redis-queue":
-        worker = RQManager(
-            config.redis_ip,
-            config.redis_port,
-            config.redis_db,
-            auth.API_key,
-            auth.session,
-        )
+        worker = RQManager(auth.API_key, auth.session, config)
```
btel, line 100: why did you remove the ip/port configuration?

quetz/testing/fixtures.py
```diff
@@ -144,6 +144,8 @@ def config_base(database_url, plugins):
 
 [plugins]
 enabled = {plugins}
+
+[redis]
```
btel, line 148: make sure that the tests also pass if [redis] section is not present, you can check that in the code using `config.configured_section("redis")`

quetz/tasks/workers.py
```diff
@@ -29,7 +29,7 @@ def prepare_arguments(func: Callable, **resources):
     return kwargs
 
 
-def wrapper(func, api_key, browser_session, **kwargs):
+def wrapper(func, api_key, browser_session, config, **kwargs):
```
btel, line 32: maybe more telling  function name and docstring would be helpful

quetz/testing/fixtures.py
btel: can you also add the extra config settings to the sphinx docs, it can be done in a separate PR

.github/workflows/ci.yml
```diff
@@ -30,6 +30,15 @@ jobs:
...
           --health-retries 5
+      redis:
+        image: redis
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 6379:6379
```
btel, line 41: I don't see where you initialize the workers and the redis server, it can be done in the CI script or in the test

btel (commented):
looks great! Just a few, hopefully helpful, comments, cheers

quetz/deps.py
madhur-tandon: because we are getting it from `config` now.

quetz/testing/fixtures.py
madhur-tandon: yeah, thanks, I didn't know that we wanted to make it pass even without the `[redis]` section. Will do the docs too.

quetz/tasks/workers.py
madhur-tandon: cool, agreed. will change.

.github/workflows/ci.yml
madhur-tandon: Yup, this was the last missing piece.

btel:
it looks good, it would be also great to add docs about setting up quetz with redis workers (launching redis, rq and workers, + configuration that we already discussed)

docs/source/deploying/configuration.rst
```diff
@@ -122,6 +122,24 @@ Quetz can store package in object cloud storage compatible with S3 interface. To
...
+``redis`` section
+^^^^^^^^^^^^^^^^^^^^^^
+
+Quetz can be configured to run with a redis-server and use redis-queue for queueing jobs and processing them in the background with workers.
+Note that you might need to run ``redis-server`` and ``rq worker`` commands in separate terminals / containers along with quetz.
+You can configure the redis-server by setting the host, port and db:
+
+.. code::
+
+   [sqlalchemy]
```
wolfv, line 134: `[redis]`?

madhur-tandon: yeah, fixed

docs/source/deploying/configuration.rst
```diff
@@ -122,6 +122,24 @@ Quetz can store package in object cloud storage compatible with S3 interface. To
 :bucket_prefix:
 :bucket_suffix: channel directories on S3 are created with the following semantics: ``{bucket_prefix}{channel_name}{bucket_suffix}``
 
+``redis`` section
+^^^^^^^^^^^^^^^^^^^^^^
+
+Quetz can be configured to run with a redis-server and use redis-queue for queueing jobs and processing them in the background with workers.
+Note that you might need to run ``redis-server`` and ``rq worker`` commands in separate terminals / containers along with quetz.
```
btel, line 129: can you create a new file in docs/source/deploying/workers.rst describing the procedure step by step, including the installation and configuration of redis (you can also pull redis from docker).

the config options should be also given  here.

docs/source/deploying/configuration.rst
```diff
@@ -122,6 +122,24 @@ Quetz can store package in object cloud storage compatible with S3 interface. To
...
+You can configure the redis-server by setting the host, port and db:
+
+.. code::
+
+   [redis]
+   ip = "127.0.0.1"
+   port = 6379
+   db = 0
```
btel, line 137: what is db? is this an integer or string? how can we make sure that the db exist?

docs/source/deploying/configuration.rst
btel: you can also mention the other worker types. If you don't have time, I can also do this after this PR is merged.

docs/source/deploying/configuration.rst
```diff
@@ -122,6 +122,24 @@ Quetz can store package in object cloud storage compatible with S3 interface. To
 :bucket_prefix:
 :bucket_suffix: channel directories on S3 are created with the following semantics: ``{bucket_prefix}{channel_name}{bucket_suffix}``
 
+``redis`` section
```
btel, line 125: I guess you will also need to document the `[workers]` section

docs/source/deploying/configuration.rst
madhur-tandon: We should do this in a separate PR.

docs/source/deploying/configuration.rst
madhur-tandon: The default database that `redis` uses is called `db` 0 to store stuff.

quetz/deps.py
```diff
@@ -92,17 +92,22 @@ def get_tasks_worker(
     else:
         worker = "thread"
 
+    if config.configured_section("redis"):
+        redis_ip = config.redis_ip
+        redis_port = config.redis_port
+        redis_db = config.redis_db
+    else:
+        redis_ip = "127.0.0.1"
```
btel, line 100: I am not sure whether these are reasonable defaults, perhaps we should force the user to provide the values if the worker is set to redis?

madhur-tandon: Yeah, we can do that too. The default `IP`, `Port`, and `db` are always those if you run a server `locally`.

quetz/tasks/workers.py
```diff
@@ -143,21 +151,38 @@ async def wait(self):
...
+            **kwargs,
         )
 
-    def wait(self):
+    async def wait(self):
         while not self.job.is_finished:
             time.sleep(1)
-        return self.job.result
+        if self.job.result:
+            return await self.job.result
```
btel, line 188: you can't await results, it's not a asynchronous function, but the result of the execution (`None` in this case).

quetz/deps.py
btel: so maybe we can keep them, but the defaults should be handled in quetz.config.Config, not here

quetz/tasks/workers.py
madhur-tandon: This is done to make it compatible with the `tests`. So that the statement `await any_worker.wait()` in `test_threading_worker_execute` and `test_threading_worker_execute_with_dao` works for any type of worker. It anyway worked for `threading` and `subprocess`. But it fails with `redis` if `async` and `await` are not used.

madhur-tandon: But I guess we can do away with `await` here. And we will still need to use `async` perhaps.

docs/source/deploying/configuration.rst
madhur-tandon: Done via https://github.com/mamba-org/quetz/pull/202

PR #199, opened 2020-12-01
get/delete endpoints for single package versions

btel:
url scheme: 

GET/DELETE `/channels/my-channel/packages/zchunk/versions/linux-64/zchunk-0.1-0.tar.bz2`

- [x] add get package version endpoint
- [x] add delete package version endpoint
- [x] check permissions
- [x] update runexports to new url scheme
- [ ] ~~extra: check permissions on listing endpoint~~ (this is nontrivial case, and shouldn't be covered in this PR)

### DIFF
--- a/plugins/quetz_runexports/quetz_runexports/api.py
+++ b/plugins/quetz_runexports/quetz_runexports/api.py
@@ -1,6 +1,6 @@
 import json
 
-from fastapi import APIRouter, Depends, HTTPException, Path, status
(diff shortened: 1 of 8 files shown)


btel:
@wolfv good to merge

wolfv:
Do you want to remove the last three commits?

btel:
If it's ok, let's leave them. There's a test I might want to reuse in a future PR.

PR #205, opened 2020-12-03
refactor and parallize package handling

wolfv:


quetz/utils.py
```diff
@@ -21,3 +22,19 @@ def add_entry_for_index(files, subdir, fname, data_bytes):
...
+class TicToc:
+    def __init__(self, description):
+        self.description = description
+
+    def __enter__(self):
+        self.start = time.time()
+
+    def __exit__(self, ty, val, tb):
+        self.stop = time.time()
+        print(f"[TOC] {self.description}: {self.stop - self.start}")
```
btel, line 36: logger.debug ?

btel (approved):
does it help to untar the files in threads?

quetz/utils.py
wolfv: yeah I tried to get a logger here, but then all logs were printed twice. Do i need to get a logger with a different name?

wolfv:
I was hoping it would help but I don't think it does that much. But uploading to S3 in threads would probably help use all the available bandwidth ... so that's what I'd like to do next.

quetz/utils.py
btel: no, I think that's a bug of uvicorn, it was fixed upstream but we need to wait for a release

wolfv:
@btel one test is still failing, and I gotta run for half an hour. If you can spot what's going on, feel free to push to this PR!

wolfv:
Going to deploy this to see what happens

PR #210, opened 2020-12-03
Delete user endpoint

btel:
rather than deleting users we remove their profiles, identities and api keys. This allows us to keep the packages and channels, which might still be shared with others.


wolfv:
Cool, finally we're GDPR compliant :)

PR #212, opened 2020-12-03
add --skip-if-exists to create and run commands

sanzoghenzo:
added this option to be able to use a single command no matter if the deployment is present or not.

With this one can:

- set a docker container with the `quetz run --skip-if-exists deployment_name` as CMD;
- place the `deployment_name` configuration in an external directory;
- run the docker with a the external directory mounted as the working directory;
- run and restart the docker without any error.

not sure if the name of the option is good, at first I thought of `--exist-ok`.

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -398,7 +398,9 @@ def create(
 
     if os.path.exists(path) and abs_path in deployments:
         if skip_if_exists:
-            logger.info(f'Quetz deployment already exists at {path}, skipping creation.')
+            logger.info(
+                f'Quetz deployment already exists at {path}, skipping creation.'
+            )
             return
         delete_ = typer.confirm(f'Quetz deployment exists at {path}.\nOverwrite it?')
         if delete_:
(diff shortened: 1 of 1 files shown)


btel:
thanks @sanzoghenzo, it looks good. I was just wondering whether your use cases could be also covered with the `quetz start` command. 

I also saw that the linter is complaining, the best way to find the linting issues locally and fix them automatically, is to run pre-commit:

```
conda install -c conda-forge pre-commit
pre-commit run --all
```

sanzoghenzo:
My bad, the funny thing is at my company I'm trying to enforce code styling, and I forgot to do that on a public project 😩 

`quetz start` doesn't initialize the db and store the deployment, so at the first run of a docker that uses that command the uvicorn server starts ok, but at the first request that needs the db it raises all sorts of errors.

Now that I think of it, `start` should add some checks before starting uvicorn...

wolfv:
Thanks!

Issue #197, opened 2020-12-01
creating Config without path raises unhandled TypeError


btel:
this only happens if `QUETZ_CONFIG_FILE` is not defined
```
import os; os.listdir()
from quetz import config
config.Config()
```

gives this error

```
['channels', 'quetz.sqlite', 'quetz.log', 'config.toml']

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bartosz/repos/libraries/mamba-org/quetz/quetz/config.py", line 140, in __new__
    path = os.path.abspath(cls.find_file(deployment_config))
  File "/home/bartosz/.pyenv/versions/micromamba/envs/quetz/lib/python3.8/posixpath.py", line 374, in abspath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

but  this works:

```
>>> os.environ['QUETZ_CONFIG_FILE'] = 'config.toml'
>>> config.Config()
<quetz.config.Config object at 0x7f8885b89610>
```

### What needs to be done

catch the exception and ask user to provide the name or path of config file using env var

### Remarks

quetz run/start configures the QUETZ_CONFIG_FILE automatically

madhur-tandon:
Should we put the call to `config.Config()` in a `try` block in all places?
Ideally this should be done only once right? And then propagated to all other places where it's needed.

btel:
no it should be done in Config class, about here:

```
File "/home/bartosz/repos/libraries/mamba-org/quetz/quetz/config.py", line 140, in __new__
    path = os.path.abspath(cls.find_file(deployment_config))
```

madhur-tandon:
yeah, thanks. This is exactly what I meant by doing it at one place - which is propagated to everywhere we call `Config()`.
I am picking this up. Will send a PR soon.

PR #214, opened 2020-12-03
ask user to set the QUETZ_CONFIG_FILE environment variable

madhur-tandon:
Fixes https://github.com/mamba-org/quetz/issues/197

### DIFF
--- a/quetz/config.py
+++ b/quetz/config.py
@@ -150,20 +150,20 @@ class Config:
             return cls._instances[None]
         try:
             path = os.path.abspath(cls.find_file(deployment_config))
-            if path not in cls._instances:
-                config = super().__new__(cls)
-                config.init(path)
-                cls._instances[path] = config
-                # optimization - for default config path we also store the instance
-                # under None key
-                if not deployment_config:
-                    cls._instances[None] = config
-            return cls._instances[path]
         except TypeError:
             raise ValueError(
                 "Environment Variable QUETZ_CONFIG_FILE \
                  should be set to name / path of the config file"
             )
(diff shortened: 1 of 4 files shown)


quetz/config.py
```diff
@@ -148,16 +148,22 @@ class Config:
...
-            config = super().__new__(cls)
-            config.init(path)
-            cls._instances[path] = config
-            # optimization - for default config path we also store the instance
-            # under None key
-            if not deployment_config:
-                cls._instances[None] = config
-        return cls._instances[path]
+        try:
+            path = os.path.abspath(cls.find_file(deployment_config))
```
btel, line 152: I think it would be enough to wrap the `try: except:` block only around this line

btel (commented):
it would be also nice to have a unit test for it. you could just try testing whether Config() raises the error if a config.toml does not exist (using `with pytest.raises(ValueError):` context manager.

quetz/config.py
madhur-tandon: That wouldn't work. Since if it fails, `path` would be undefined and then the rest of the code block wouldn't work.

btel (commented):
@madhur-tandon I pushed a commit to fix the test on ci (I hope it works this time). the problem was with a fixture that did not clean up properly after other tests run.

quetz/config.py
madhur-tandon: Thanks, somehow it didn't work earlier, but now it does. Probably because I wasn't in the `deployment` directory. I have pushed the changes.

btel (approved):
well done! :+1:

PR #220, opened 2020-12-07
Case insensitive channel names

### LINKED ISSUE #163: channel names should be case insensitive
Currently if I have a `RoboStack` channel, it's different from `robostack`. However, I think (also for security reasons) that those names should be case insensitive and both should point to the same channel.
Same for user names etc.

btel:
fixes #163


quetz/tests/api/test_channels.py
```diff
@@ -232,3 +232,42 @@ def test_unique_channel_names_are_case_insensitive(auth_client, user, db):
...
+
+    response = auth_client.get(f"/api/channels/{channel_name}")
+
+    assert response.status_code == 200
+    assert response.json()['name'] == channel_name
+
+
+def test_accents_make_unique_channel_names(auth_client, maintainer):
+
+    channel_names = ["żmija", "zmija", "grün", "grun"]
```
wolfv, line 261: does this mean we allow all 4 channel names to exist?

This also might allow for vicious attacks. Unicode has many characters that look similar.

For example: c⃝nda-f⃝rge

wolfv: ok this example doesn't work but you get the idea :)

PR #222, opened 2020-12-07
remove deployments.json from CLI

### LINKED ISSUE #216: cli "deployments" rationale (and _user_dir breaking docker)
**TL;DR**: if I run `quetz run` in a docker container without a user directory, the container exits with a permission error while trying to create the directory `/nonexistent/.config/quetz` in which to place the `deployments,json` file.
I'm wondering if this config file is needed at all. 

This is my Dockerfile, that builds the latest git version of quetz and runs it with the option added with #212:

```dockerfile
FROM condaforge/miniforge3:latest as base

WORKDIR build
ARG CONDA_ENV=/opt/venv
COPY environment.yml .
RUN conda env create -p $CONDA_ENV -f environment.yml
COPY pyproject.toml setup.py LICENSE ./
COPY quetz ./quetz
RUN conda run -p $CONDA_ENV pip install . --no-deps

FROM gcr.io/distroless/base-debian10 as production
ARG CONDA_ENV=/opt/venv
ENV LANG=en_US.UTF-8
ENV PATH $CONDA_ENV/bin:$PATH
COPY --from=base $CONDA_ENV $CONDA_ENV
USER nobody
WORKDIR /deployment
ENTRYPOINT ["quetz"]
CMD ["run", ".", "--skip-if-exists"]
```

as you can see the working directory is set to `/deployment` and the default command tries to run the deployment in the current directory, if it finds a `config.toml` in it.

with a `docker-compose.yml` like this:
```yml
version: "3.8"

services:
  quetz:
    image: "quetz:latest"
    build:
      context: ..
      dockerfile: docker/Dockerfile
    container_name: quetz
    ports:
      - "8000:8000"
    volumes:
      - "./deployment:/deployment"
```
I can mount a local deployment directory with the `config.toml` in it.

the aformentioned error is originated by the[`_get_deployments` function](https://github.com/mamba-org/quetz/blob/master/quetz/cli.py#L264) that tries to create the directory in which to place the config file.

this file is only used to keep track of existing deployments, but I don't think it's useful in a docker context;
I suppose that the (non)existence check can be better performed by looking at the actual files and directories inside the deployment folder.

If it's ok with you I'm going to try to fix the issue by removing this file.

Maybe we should try to better define the cli commands, but I think it's already mentioned in #155

sanzoghenzo:
This will replace the functions related to `deployments.json`
with a function that checks if a specified path is a deployment, that is:
- has a 'config.toml' file
- has a 'channels' directory
- the database specified in the config file exists.

The database existence check is performed via sqlalchemy_utils.

`quetz create` has a new `delete` option instead of the delete confirmation,
and only works if either `copy_conf` or `create_conf` is specified.
`skip_if_exists` is renamed to `exists_ok`, also.

`config_file_name` is removed from both `run` and `create` commands,
`config.toml` is the only allowed config file name.
file specified in `copy_conf` will be renamed to `config.toml` in the deployment directory.

Pathlib objects are used extensively.
we also use a context manager `working_directory`
to ensure we return to the original folder after the run.

closes #216

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -1,7 +1,6 @@
 # Copyright 2020 QuantStack
 # Distributed under the terms of the Modified BSD License.
-
-import json
+import contextlib
 import logging
 import os
 import random
@@ -11,7 +10,7 @@ import uuid
 from distutils.spawn import find_executable
 from enum import Enum
 from pathlib import Path
-from typing import Dict, NoReturn, Optional
+from typing import NoReturn, Optional, Union
 
 import pkg_resources
 import typer
@@ -19,12 +18,12 @@ import uvicorn
 from alembic import command
 from alembic.config import Config as AlembicConfig
 from sqlalchemy.orm.session import Session
+from sqlalchemy_utils.functions import database_exists
 
 from quetz.config import (
     Config,
     _env_config_file,
     _env_prefix,
-    _user_dir,
     configure_logger,
     create_config,
 )
@@ -43,8 +42,6 @@ from quetz.db_models import (
 
 app = typer.Typer()
 
-_deployments_file = os.path.join(_user_dir, 'deployments.json')
-
 logger = logging.getLogger("quetz-cli")
 configure_logger(loggers=("quetz-cli", "alembic"))
 
@@ -56,6 +53,17 @@ class LogLevel(str, Enum):
     info = "info"
     debug = "debug"
     trace = "trace"
+
+
+@contextlib.contextmanager
+def working_directory(path):
+    """Change working directory and return to previous on exit."""
+    prev_cwd = Path.cwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(prev_cwd)
 
 
 def _alembic_config(db_url: str) -> AlembicConfig:
@@ -105,7 +113,7 @@ def _make_migrations(
         if entry_point.name == plugin_name:
             found = True
 
-    if not plugin_name == "quetz" and not found:
+    if plugin_name != "quetz" and not found:
         raise Exception(
             f"models entrypoint (quetz.models) for plugin {plugin_name} not registered"
         )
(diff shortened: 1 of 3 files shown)


quetz/cli.py
```diff
@@ -249,76 +258,17 @@ def _fill_test_database(db: Session) -> NoReturn:
...
+def _is_deployment(base_dir: Path):
+    config_file = base_dir.joinpath("config.toml")
+    if (
+        base_dir.exists()
+        and config_file.exists()
+        and base_dir.joinpath("channels").exists()
+    ):
+        config = Config(str(config_file.resolve()))
+        with working_directory(base_dir):
+            return database_exists(config.sqlalchemy_database_url)
```
btel, line 270: this may be misleading if the user has configured the pg db url, but did not create the database yet (or had forgotten to do so).

sanzoghenzo: you're right.
What are the minimal checks I should perform?

btel: maybe we could just skip this check and raise an error at runtime if the db does not exist. What do you think @wolfv?

quetz/tests/test_cli.py
```diff
@@ -86,9 +86,8 @@ def get_db(_):
     with mock.patch("quetz.cli.get_session", get_db):
         cli.create(
             Path(config_dir) / "new-deployment",
-            config_file_name="config.toml",
             copy_conf="config.toml",
-            create_conf=None,
+            create_conf=False,
             dev=True,
         )
```
btel, line 93: it would be cool to add some unit tests. cli module is one of the least tested, according to codecov: https://codecov.io/gh/mamba-org/quetz/tree/master/quetz

quetz/cli.py
wolfv: I agree, a runtime error is fine with me

wolfv:
thanks!

sanzoghenzo:
😱  I didn't fully resolve your reviews yet!
I'll add some more tests and fix the `_is_deployment` checks in another MR then 😉

wolfv:
great! yeah I was deploying quetz, and something broke my docker-compose usage. And then this PR fixed it, hehe.

PR #226, opened 2020-12-08
add github style search for Packages

madhur-tandon:
Linting would probably fail because the regex is long. And changing it so that it is multiple lines kind of changes the regex if I recall correctly. So this can be ignored for now.

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -845,7 +845,7 @@ def delete_package_version(
 
 
 @api_router.get(
-    "/search/", response_model=List[rest_models.PackageSearch], tags=["search"]
+    "/packages/search/", response_model=List[rest_models.PackageSearch], tags=["search"]
 )
 def search(
     q: str,
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -68,7 +68,7 @@ from quetz.rest_models import ChannelActionEnum, CPRole
 from quetz.tasks import indexing
 from quetz.tasks.common import Task
 from quetz.tasks.mirror import LocalCache, RemoteRepository, get_from_cache_or_download
-from quetz.utils import TicToc, generate_random_key
+from quetz.utils import TicToc, generate_random_key, parse_query
 
 from .condainfo import CondaInfo
 
@@ -845,15 +845,16 @@ def delete_package_version(
 
 
 @api_router.get(
-    "/search/{query}", response_model=List[rest_models.PackageSearch], tags=["search"]
+    "/search/", response_model=List[rest_models.PackageSearch], tags=["search"]
 )
 def search(
-    query: str,
+    q: str,
     dao: Dao = Depends(get_dao),
     auth: authorization.Rules = Depends(get_rules),
 ):
     user_id = auth.get_user()
-    return dao.search_packages(query, user_id)
+    keywords, filters = parse_query(q)
+    return dao.search_packages(keywords, filters, user_id)
 
 
 @api_router.get("/api-keys", response_model=List[rest_models.ApiKey], tags=["API keys"])
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -853,8 +853,21 @@ def search(
     auth: authorization.Rules = Depends(get_rules),
 ):
     user_id = auth.get_user()
-    keywords, filters = parse_query(q)
+    keywords, filters = parse_query('package', q)
     return dao.search_packages(keywords, filters, user_id)
+
+
+@api_router.get(
+    "/channels/search/", response_model=List[rest_models.ChannelSearch], tags=["search"]
+)
+def channel_search(
+    q: str,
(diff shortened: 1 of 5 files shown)


quetz/utils.py
```diff
@@ -24,6 +30,74 @@ def add_entry_for_index(files, subdir, fname, data_bytes):
     )
 
 
+def parse_query(query):
+    accepted_filters = ['channel', 'description', 'summary']
+    query = unquote(query.strip())
```
btel, line 35: do you need it? I would assume that fastapi would already do it for you

quetz/utils.py
```diff
@@ -24,6 +30,74 @@ def add_entry_for_index(files, subdir, fname, data_bytes):
...
+    regex = r"""(\S+:'(?:[^'\\]|\\.)*')|(\S+:\"(?:[^\"\\]|\\.)*\")|(-?\"(?:[^\"\\]|\\.)*\")|(-?'(?:[^'\\]|\\.)*')|\S+|\S+:\S+"""
```
btel, line 36: impressive! perhaps you might use named groups to make it more legible

also make sure that, you can skip any of the query terms. For example, you might search only by channel name:

```
channel=conda-forge,my-channel
```

it's possible that you already do this, i don't know

quetz/main.py
```diff
@@ -744,15 +744,16 @@ def delete_package_version(
...
 def search(
-    query: str,
+    q: str,
     dao: Dao = Depends(get_dao),
     auth: authorization.Rules = Depends(get_rules),
 ):
     user_id = auth.get_user()
-    return dao.search_packages(query, user_id)
+    keywords, filters = parse_query(q)
+    return dao.search_packages(keywords, filters, user_id)
```
btel, line 756: we should also make sure that the user has permission to see the packages from the search results (based on channel and package permission)

quetz/utils.py
```diff
@@ -24,6 +30,74 @@ def add_entry_for_index(files, subdir, fname, data_bytes):
     )
 
 
+def parse_query(query):
+    accepted_filters = ['channel', 'description', 'summary']
```
btel, line 34: how about filtering by user (owner)?

btel (commented):
it looks very good. make sure that you cover all possible variations in the tests (see above for one example). Another thing to check is whether you can add arbitary number of whitespace between the query terms. This should give the same result:

`channel=conda-forge description=' test package'`

`channel=conda-forge      description=' test package'`

also whitespaces at the beginning and end should be allowed. 

I think that using a parser might simplify a bit the regex stuff. Have a look at the parser that I sent you (https://ply.readthedocs.io/en/latest/ply.html#lex-example) or different one, and check whether it might help. At the end, you are constructing here a DSL (domain-specific language,  Also have a look at this blog post: https://medium.com/swlh/how-to-write-your-own-domain-specific-language-in-python-a36b72d3e36c

quetz/utils.py
wolfv: or uploader. although that is more tricky as that is part of the packageversion data.

quetz/utils.py
wolfv: yeah a bit more documentation on this regex would be great :)

also maybe we could write a small parser for this mini-language, that's not based on regex?

quetz/utils.py
madhur-tandon: Yes, we are mimicking of what we would ideally receive from text entered into a search box. We can anyway remove this later if we figure out that FastAPI does this OR perhaps if the content entered into a search box stays the same.

quetz/utils.py
madhur-tandon: Yes, it is not necessary to supply all the arguments. It's up to us how many we supply. I did give a demo of this today.
I partially borrowed this regex (and did some of my own modifications) so I will have to see how it works at a deeper level as well.

I can check docs about named groups. And, I need to investigate about making a parser a bit more.

quetz/utils.py
madhur-tandon: Yes, we are gonna expand functionality one by one. Right now, all things listed in `PackageSearch` class in `rest_models.py` are considered.

quetz/utils.py
wolfv: @madhur-tandon maybe using the Python tokenizer is even easier? It splits out "strings" with ticks already, and it's part of the standard library.

https://docs.python.org/3/library/tokenize.html

I am sure there are other tokenizer libraries implemented in Python around. That might be easier & more maintainable than this regex, in the long term :)

madhur-tandon: We are now using `shlex` (https://docs.python.org/3/library/shlex.html) as suggested by @wolfv

quetz/utils.py
madhur-tandon: Added `uploader` as well

PR #229, opened 2020-12-09
make github login optional, frontend login buttons

sanzoghenzo:
- the github part of the config is now optional;
- two new test endpoint to check if the google/github auth is enabled;
- Vue frontend checks for these endpoints to show the login buttons;
- using the router prefix instead of declaring the full path for each entrypoint
- minor refactors

I'm sorry I didn't add unit tests, but I cannot wrap my head around the config object...  How do I create a test config with the google entries (and without the github ones)?


wolfv:
Great! I wanted this for a while :)

PR #230, opened 2020-12-10
add more channel_actions

wolfv:


quetz/tasks/common.py
```diff
@@ -14,6 +14,11 @@
 def assert_channel_action(action, channel):
     if action == ChannelActionEnum.synchronize:
         action_allowed = assertions.can_channel_synchronize(channel)
+    elif action == ChannelActionEnum.validate_packages:
+        # TODO!!!
+        action_allowed = True
+    elif action == ChannelActionEnum.generate_indexes:
+        action_allowed = True
```
btel, line 21: I guess we don't want everyone to trigger these actions, it would be an easy way to DoS the server

quetz/tasks/mirror.py
```diff
@@ -410,7 +410,7 @@ def synchronize_packages(
     if subdirs is None:
         subdirs = KNOWN_SUBDIRS
 
-    for arch in subdirs:
+    for arch in ['noarch']:
```
btel, line 413: ??? I think this causes the CI to fail

wolfv: I was using this branch to deploy and test on the server, sorry. Wasn't really ready for review

quetz/tasks/common.py
wolfv: I think here we're only testing if the channel itself can do that (mirror vs real channel etc).

quetz/main.py
```diff
@@ -1052,7 +1024,12 @@ def handle_package_files(
         if force:
             auth.assert_overwrite_package_version(channel_name, package_name)
 
-        logger.debug(f"adding file '{file.filename}' to channel '{channel_name}'")
+        file.file._file.seek(0, os.SEEK_END)
```
btel, line 1027: I would avoid accessing the _file attribute, it's sort of hack that only works with SpooledTemporaryFile and not a standard file-interface (it's a known issue: https://github.com/python/cpython/pull/3249). You can instead patch the SpooledTemporaryFile as done in CondaInfo: https://github.com/mamba-org/quetz/blob/e552b8afff6df10ceb36c30e5ced47e14de58d44/quetz/condainfo.py#L160-L162

quetz/main.py
```diff
@@ -1052,7 +1024,12 @@ def handle_package_files(
...
+        size = file.file._file.tell()
+        total_size += size
```
btel, line 1029: for the channel size limit: do we want to check whether all the files can be uploaded or we do that one by one and fail after N files have been already uploaded

quetz/main.py
wolfv: Done

quetz/main.py
wolfv: I like to fail early instead of halfway through.

PR #240, opened 2020-12-14
adding query syntax

btel:
Needs some extra testing!

Example package spec:

```
"my-package>=0.1.2,<0.2,other-package==1.1"
```

which will run the job for packages:
* `my-package` with version between 0.1.2 and 0.2
* `other-package` in version 1.1

Note OR (`|`) not yet supported.

### Reference implementations
* conda:
- MatchSpec (docstring): https://github.com/conda/conda/blob/aed799b56d9ee16a3653928527e2af6e41394e85/conda/models/match_spec.py#L73
- parsing: https://github.com/conda/conda/blob/aed799b56d9ee16a3653928527e2af6e41394e85/conda/models/match_spec.py#L569
- conda docs: https://github.com/conda/conda/blob/aed799b56d9ee16a3653928527e2af6e41394e85/docs/source/user-guide/concepts/pkg-specs.rst#package-match-specifications

* PEP440: https://www.python.org/dev/peps/pep-0440/#version-scheme

### DIFF
--- a/plugins/quetz_transmutation/quetz_transmutation/api.py
+++ b/plugins/quetz_transmutation/quetz_transmutation/api.py
@@ -12,6 +12,8 @@ from quetz.deps import get_db, get_rules
 from quetz.jobs.models import Job
 from quetz.pkgstores import PackageStore
 
+from .rest_models import PackageSpec
+
 router = APIRouter()
 
 
@@ -59,11 +61,17 @@ def transmutation(package_version: dict, config, pkgstore: PackageStore, dao: Da
 transmutation_serialized = pickle.dumps(transmutation)
 
 
-@router.get("/api/transmutation")
+@router.put("/api/transmutation")
 def put_transmutation(
-    db=Depends(get_db), auth: authorization.Rules = Depends(get_rules)
+    package_spec: PackageSpec,
+    db=Depends(get_db),
+    auth: authorization.Rules = Depends(get_rules),
 ):
     user = auth.get_user()
-    job = Job(owner_id=user, manifest=transmutation_serialized)
+    job = Job(
+        owner_id=user,
+        manifest=transmutation_serialized,
+        items_spec=package_spec.package_spec,
+    )
     db.add(job)
     db.commit()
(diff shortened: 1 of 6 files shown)


wolfv:
Cool! I'll merge, but I'm afraid version comparison might not fully work. For example, when you have 3 versions like 

0.2.0
0.12.0
1.0.0

I'm not sure how string comparison is implemented in sql.

We might actually have to add the python function as extension to Postgres :)

btel:
and how about defining the version as a composite type (major, minor, patch) and using this trick? :


https://docs-sqlalchemy.readthedocs.io/ko/latest/orm/composites.html#redefining-comparison-operations-for-composites

wolfv:
yeah, i guess that would work, but also not for all cases. I think really the best & coolest would be to figure out how to properly integrate the spec into postgres :)

We could have a postgres extension, and a Python fallback.

plugins/quetz_transmutation/quetz_transmutation/api.py
```diff
@@ -12,6 +12,8 @@
 from quetz.jobs.models import Job
 from quetz.pkgstores import PackageStore
 
+from .rest_models import PackageSpec
```
wolfv, line 15: @btel I think this file is missing?

dhirschfeld:
This is a neat trick:
```sql
SELECT id, version
FROM   versions
ORDER  BY string_to_array(version, '.')::int[];
```
https://dba.stackexchange.com/a/74298

There's also an existing `semver` extension for postgres:
https://pgxn.org/dist/semver/doc/semver.html

*Disclaimer:* I haven't used it so YMMV! Also, PEP-440 is a bit different from SemVer IIUC

plugins/quetz_transmutation/quetz_transmutation/api.py
btel: ah, sorry, I was tired. but why did the tests pass 🤔 ?!

btel:
I would prefer to avoid using pg extensions. It complicates the installation and breaks compatibility with other databases. 

In this case we could come up with a neat sql query to do the trick.

PR #248, opened 2020-12-15
fix package spec without version

btel:


wolfv:
yep, I hit that, too :) cool, thanks!

PR #249, opened 2020-12-15
add catch-all router for single page app

wolfv:


wolfv:
@hbcarlos this should work for what we discussed. routing can happen easily in the frontend now :)

hbcarlos:
Nice! Thanks Wolf!!

PR #253, opened 2020-12-16
add api for jobs & tasks

wolfv:


quetz/jobs/models.py
```diff
@@ -62,6 +65,8 @@ class Job(Base):
         default=JobStatus.pending,
     )
 
+    tasks = sa.orm.relationship('Task', cascade="all,delete-orphan")
```
wolfv, line 68: @btel is this fine? I changed the relationship / backref stuff to appease mypy.

btel: yes, it should be the same ;-)

btel (commented):
cool!

PR #255, opened 2020-12-16
write to tempdir before moving repodata to final destination

wolfv:
This should make our repodata updates safer, and also make sure that we don't have unpatched repodata in the final destination. 

@btel do you have any remarks?


btel (approved):
lgtm

PR #258, opened 2020-12-17
Add download counts

btel:


quetz/dao.py
```diff
@@ -677,3 +678,86 @@ def get_jobs(self):
...
+                    .filter(PackageVersion.platform == platform)
+                ).one()
+                m = PackageVersionMetric(
+                    package_version=package_version,
+                    metric_name=metric_name,
+                    period=interval,
+                    timestamp=now_interval,
+                )
+                self.db.add(m)
+                self.db.flush()
```
wolfv, line 732: do we need to flush here?

btel: yes, otherwise incrementing the count wouldn't work (`m.count` would be None). flush does not yet commit, only synchronises with the database engine

btel:
🎉

PR #274, opened 2020-12-23
add ability to whitelist / blacklist packages while creating a mirror

madhur-tandon:


quetz/main.py
```diff
@@ -542,6 +542,24 @@ def post_channel(
...
+        )
+
+    package_list = []
+    include = False
+    if includelist is not None:
+        package_list = includelist
+        include = True
+    if excludelist is not None:
+        package_list = excludelist
+        include = False
```
wolfv, line 561: wouldn't it be better to keep `includelist` and `excludelist`? I think the API would be cleaner that way

quetz/tasks/mirror.py
```diff
@@ -357,24 +361,28 @@ def handle_batch(update_batch):
             return False
 
         for package_name, metadata in packages.items():
-            path = os.path.join(arch, package_name)
+            if (include and check_package_membership(package_name, package_list)) or (
```
wolfv, line 364: this if condition is really confusing. Is there a way to simplify?

wolfv (changes requested):
Would be cool to make the API more obvious and simplify the if condition.

quetz/tasks/mirror.py
madhur-tandon: I actually put it inside a KMAP solver / boolean algebra calculator. It cannot be simplified further. It's basically the condition `(A and B) or (!A and !B)` where A is `include` and B is `check_package_membership`. This intelligently takes care of both `inclusion` and `exclusion` because depending on `include` as True/False,
we check the first term `(A and B) => B because A is already true`
OR the second term `(!A and !B) => !B because A was false and thus !A is true`.

At the end, we check either `B` (include) or `!B` (exclude) based on the state of `A` (which tells us what to do).
So for each package, we either enter the loop or do not enter it (based on `B` or `!B`)

quetz/main.py
madhur-tandon: Hmm, then we would have to pass these 2 variables (`includelist` and `excludelist`). And then, we need to check which one is empty and which one isn't (because one of them will always be empty else we throw `422`). 

So passing these 2 variables where one is kind of just deadweight to the worker (through `**kwargs`) seemed a bit dull to me.

And then, based on which one is not empty, we would need to call `check_package_membership` on the one which is not empty. This kind of thing would again lead us to the `if` statement such as:

`if ((includelist!=[] and check_package_membership(package_name, includelist)) or (excludelist!=[] and (not check_package_membership(package_name, excludelist)))`

This is even more complicated than what we have IMO.

Plus, there is this issue of default values. With our current implementation, we can supply `include` as `False` and `package_list` as `[]` which is essentially equivalent to `excludelist=[]` or saying that `exclude nothing` aka `include everything`.

But with 2 separate lists, what could be the default value for `includelist`? We wouldn't want to fetch the data to calculate this.

quetz/tasks/mirror.py
wolfv: One way to simplify difficult if conditions is by splitting and using `continue` or `break`. Maybe that works here?

quetz/main.py
wolfv: In Python, an empty list is "falsy" so `includelist and check_package_membership(name, includelist)` isn't so bad.

Also one could move the "empty" check into the `check_package_membership` function, then it would be only one function call.

The condition would be:

`if check_package_membership(package_name, include_list) or not check_package_membership(package_name, exclude_list):` which looks like a huge improvement in terms of readability to me :)

madhur-tandon: This condition unfortunately doesn't work. I have implemented it a bit differently. Hope that works.

madhur-tandon:
@wolfv should be ready to merge (apart from the external dependency on @btel's channel)

wolfv:
hey, this looks sooo much cleaner! awesome changes, thanks!

PR #275, opened 2020-12-23
fill db from channeldata/repodata

btel:
use `synchronize_repodata` action to trigger synchronisation based on repodata (no extraction of tarballs required). it's also a default when creating a new mirror channel.


btel:
good to merge @wolfv

PR #277, opened 2021-01-04
remove flashing user

madhur-tandon:
Python backend to remove flashing while logging.

### DIFF
--- a/quetz/frontend.py
+++ b/quetz/frontend.py
@@ -2,7 +2,6 @@ import json
 import logging
 import os
 import sys
-import uuid
 from pathlib import Path
 
 import jinja2
@@ -24,10 +23,28 @@ catchall_router = APIRouter()
 
 mock_settings_dict = None
 frontend_dir = ""
-static_dir = ""
-profile = None
-config_data = {}
 index_template = None
+
+
+google_login_available = (
+    hasattr(config, 'google_client_id')
+    and config.google_client_id is not None
+    and hasattr(config, 'google_client_secret')
+    and config.google_client_secret is not None
+)
+github_login_available = (
+    hasattr(config, 'github_client_id')
+    and config.github_client_id is not None
+    and hasattr(config, 'github_client_secret')
+    and config.github_client_secret is not None
+)
+
+config_data = {
+    "appName": "Quetz – the fast conda package server!",
+    "baseUrl": "/jlabmock/",
+    "github_login_available": github_login_available,
+    "google_login_available": google_login_available,
+}
 
 
 @mock_router.get('/api/sessions', include_in_schema=False)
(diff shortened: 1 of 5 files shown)


quetz/frontend.py
```diff
@@ -102,13 +110,33 @@ def render_index(config):
...
+    get_rendered_index(config_data, profile, index_template)
+
     if "." not in resource:
         return FileResponse(path=os.path.join(frontend_dir, "index.html"))
     else:
         return FileResponse(path=os.path.join(frontend_dir, resource))
 
 
+def get_rendered_index(config_data, profile, index_template):
+    config_data["logged_in_user_profile"] = rest_models.Profile.from_orm(profile).json()
```
wolfv, line 133: maybe it should be `dict` here? since it is going to be converted to json in the template / jinja.

quetz/frontend.py
```diff
@@ -102,13 +110,33 @@ def render_index(config):
...
+    logger.info(f"Page config: {config_data}")
+    index_rendered = index_template.render(page_config=config_data)
+    with open(static_dir / "index.html", "w") as fo:
```
wolfv, line 136: this is writing out the index.html, but it changes per user. we need to return it with a different response type (HTMLResponse).

quetz/frontend.py
```diff
@@ -102,13 +110,33 @@ def render_index(config):
...
+    session: dict = Depends(get_session),
+    dao: Dao = Depends(get_dao),
+    auth: authorization.Rules = Depends(get_rules),
+):
+    user_id = auth.assert_user()
+    global profile
+    profile = dao.get_profile(user_id)
+    profile.user.id = str(uuid.UUID(bytes=profile.user.id))
+
+    get_rendered_index(config_data, profile, index_template)
```
wolfv, line 124: we should also make sure that the other frontends continue to work fine

quetz/frontend.py
madhur-tandon: Nope, with `dict`, the `id` is `UUID('09abd3ed-2ecd-4741-9483-7b2bb23a63a7')` which is an object. This leads to `TypeError: Object of type UUID is not JSON serializable`.
But with `json`, the value is `"09abd3ed-2ecd-4741-9483-7b2bb23a63a7"`

madhur-tandon:
@btel The CI for this PR fails for the test in the `conda-suggest` plugin.
Specifically, it's in `response = client.get("/api/dummylogin/madhurt”)` -->
`return self.db.query(Profile).filter(Profile.user_id == user_id).one()` -->
`sqlalchemy.orm.exc.NoResultFound: No row was found for one()`

This seems unrelated to the contents changed in this PR since we don't even touch the plugins. Can you have a look at this if possible?

Issue #95, opened 2020-09-08
Per-channel TTL for the cache data


wolfv:
We should have a per-channel TTL for the cache data so that the server returns the appropriate data in the headers.

These are the headers considered by mamba & conda (mod and ETag)

https://github.com/mamba-org/mamba/blob/276e5a5ca2f394b240a7f166ebfa6e1a1849bbe1/src/subdirdata.cpp#L346-L352

wolfv:
For this we need to have an additional database field with the value and return it as the Cache-Max-Age header. cc @madhur-tandon

PR #286, opened 2021-01-07
add per channel TTL

madhur-tandon:
Closes https://github.com/mamba-org/quetz/issues/95


wolfv:
can you check that the migration adds the default value (10 hours) to existing rows?
We had this issue previously IIRC (cc @btel)

btel:
yes, it might be an issue here

madhur-tandon:
I have added `server_default` in `migration` and use `default` in `db_models`?
Should both of them use `server_default` or only one of them? If only one, where is it preferable?

Issue #101, opened 2020-09-10
Create channel pkgstore folder and empty noarch on creation


wolfv:
@nowster do you have an opinion on wether it would be a good idea to create the pkgstore folder / bucket and initialize the noarch channel repodata (empty) upon channel creation?

PR #289, opened 2021-01-07
create channel in pkgstore upon creation

madhur-tandon:
Fixes https://github.com/mamba-org/quetz/issues/101


wolfv:
this will make the initial user experience much better!

Issue #290, opened 2021-01-08
Maximum channel size is in bytes


wolfv:
I think the maximum channel size is currently in bytes. If we are using a regular integer, that limits the maximum allowed channel size to 2 Gigabyte which is not that much. 

We could decide to interpret the number as Megabyte.

What do you think @btel?

btel:
😄 yes, that might be an issue. good catch!

wolfv:
@madhur-tandon do you want to have a stab at this?

madhur-tandon:
The only thing is, do we want to constraint things to be as an Integer then?

btel:
isn't it already the case? the pydantic model and the db model should only accept integers

madhur-tandon:
yes, but, we would want to change that to float right?
because we divide by 1e6 to get Megabytes

btel:
no, I don't think so. 1MB of resolution should be enough.

PR #296, opened 2021-01-08
use BigInteger for sizes

madhur-tandon:
Fixes https://github.com/mamba-org/quetz/issues/290


madhur-tandon:
Migrations are not applied as of now.
Also, we don't need to change `rest_models` since an `int` in `Python 3` are of unlimited size.

btel (approved):
great work!

wolfv:
looks great! just squash 2 or 3 commits and it's ready to merge :)

PR #306, opened 2021-01-14
fix creating users with multiple providers

### LINKED ISSUE #292: specify provider when adding users in config.toml
in config.toml automatically created users are identfied by their username, which could be exploited if the application is configured with more than one provider (for example, github and google).

Steps:

- quetz uses two identity providers to connect users: github oauth and google
- admin configures github user "alice" as maintainer in config.toml ([users] section),

```
[users]
maintainers = ["alice"]
```

- alice does not connect her account
- an attacker creates an account "alice" on google
- attacker uses the google account to connect with quetz
- attacker obtains "maintainer" role

btel:
### Breaking changes

*  provider has to be specified with each user in config `[users]` section, for example:

```
[users]
owner = ["github:username"]
```

* to assign roles defined in the config, you will need to re-run `quetz init-db` **after** the user has associated their account with quetz

* only a single identity is allowed to be associated with a username, more identities can be associated with the same user by changing directly data in the db, but due to security concerns this is not exposed with the api

closes #292

### DIFF
--- a/dev_config.toml
+++ b/dev_config.toml
@@ -16,4 +16,4 @@ level = "DEBUG"
 file = "quetz.log"
 
 [users]
-admins = ["alice"]
+admins = ["dummy:alice"]
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -42,22 +42,25 @@ Configure default user permissions, creating default channel and super-admin per
 
    [users]
    # users with owner role
-   admins = ["admin_user"]
+   admins = ["github:admin_user"]
    # users with maintainer role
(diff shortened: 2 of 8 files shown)


wolfv:
Can we handle the role-assignment dynamically upon user-creation? Shouldn't we have access to the default roles from config?

btel:
we read the default role for any user from the  config.  It's only for pre-defined users that we have to create them in the db.  This is necessary, because config is loaded only once after we start the application and any changes in the config would not be available until we restart. I figured out it is faster and easier to run init-db on running server than to restart the server ;-)

PR #309, opened 2021-01-18
return roles=null for api keys with user-level permissions

wolfv:


wolfv:
@btel if you have a sec for review, that would be great :)

quetz/dao.py
```diff
@@ -409,16 +409,28 @@ def create_package_member(self, channel_name, package_name, new_member):
...
+            .filter(~ApiKey.deleted)
+            .all()
+        )
+
+        custom_role_api_keys = (
             self.db.query(ApiKey, PackageMember, ChannelMember)
             .select_from(ApiKey)
             .filter(ApiKey.owner_id == user_id)
             .filter(~ApiKey.deleted)
+            .filter(ApiKey.owner_id != ApiKey.user_id)
```
btel, line 426: probably that's not enough, because a user could create an api key for a different user. Altough at the moment you can do that with web API.

you could instead check whether a profile is not null with something like this (probably it would need some tweaking):

```
...
.outerjoin(Profile, ApiKey.user_id == Profile.user_id).filter(Profile.user_id != None)
...
```

quetz/dao.py
```diff
@@ -409,16 +409,28 @@ def create_package_member(self, channel_name, package_name, new_member):
         self.db.commit()
 
     def get_api_keys_with_members(self, user_id):
-        return (
+        user_role_api_keys = (
+            self.db.query(ApiKey)
+            .select_from(ApiKey)
```
btel, line 414: no need for `.select_from` here

quetz/dao.py
```diff
@@ -409,16 +409,29 @@ def create_package_member(self, channel_name, package_name, new_member):
         self.db.commit()
 
     def get_api_keys_with_members(self, user_id):
-        return (
+        user_role_api_keys = (
+            self.db.query(ApiKey)
+            .filter(ApiKey.owner_id == user_id)
+            .outerjoin(Profile, ApiKey.user_id == Profile.user_id)
```
btel, line 415: probably we don't need to check profile here, because only a real user can be an owner ;-)

quetz/tests/api/test_api_keys.py
```diff
@@ -81,20 +81,8 @@ def test_list_keys_with_channel_roles(
...
-            "channel": private_channel.name,
-            "package": None,
-            "role": "maintainer",
-        }
-    ]
+    assert returned_keys["user_is_user"] is None
+    assert returned_keys["key"] is None
 
 
 def test_list_keys_with_package_roles(
```
btel, line 88: this test is basically the same as the one before, also we don't need to add PackageMember/ChannelMember because we don't return the memberships with the api key any more

quetz/dao.py
wolfv: but we need the inverse of the check above otherwise we'd get API keys twice, right?

btel: I may be wrong, but this should be already covered in the condition: `APIKey.user_id == user_id` because `user_id` is an owner, so it must be a real user with a profile. but maybe it's to fragile to depend on. someone might come and make some owners fake users (for example, a team could be an owner), and this will fail. So it might be better to check the profile.

in the future, we could also add a flag to a user model, to check whether it's a real or fake user (such as `User.real = True/False`)

wolfv: I think you're right! Thnaks

PR #318, opened 2021-01-25
api_key create_at and expire_at

hbcarlos:


wolfv:
sorry, I just killed beta.mamba.pm so the tests are not working. I am bringing it up ... but it takes a little while longer :see_no_evil:

hbcarlos:
Don't worry, I'have work on the frontend. : )

wolfv:
Testing with server_default .. let's see!

hbcarlos:
I have to change the `datetime` to `date`.

wolfv:
we can also keep datetime and just put the clock to 00:00:00

wolfv:
@hbcarlos when you make the change, can you also modify to `time_created`? because that's the word we used in that other case.

hbcarlos:
Sure

wolfv:
can you squash the commits regarding linting a bit? you can also squash my two commits into one. 

I'll merge, but we need to keep in mind that we're not yet enforcing the expiration date anywhere.

PR #319, opened 2021-01-25
add download count and other metrics

madhur-tandon:


quetz/main.py
```diff
@@ -84,6 +86,8 @@
     https_only=config.session_https_only,
 )
 
+app.add_middleware(PrometheusMiddleware)
```
wolfv, line 89: Can we do this the other way around? Have a function in quetz.metrics that is called `init` and takes an `app` as input and installs the metrics routes and middleware?

e.g..

```
from quetz import metrics
...

metrics.init(app)
```

wolfv:
small nitpicky comment, otherwise all is great

quetz/main.py
madhur-tandon: Done

PR #322, opened 2021-01-27
increment counter inside endpoint

madhur-tandon:


wolfv:
looks good! I wonder if we need to make it possible to disable the metrics collection (maybe some people don't want it  ... )

But good for now.

madhur-tandon:
we can add a command line argument that disables or enables `metrics.init(app)` then?

wolfv:
sure. but now we're also "invading" other source files which will make it harder to disable the metrics.

PR #324, opened 2021-01-27
optimizations for download endpoint

btel:
optimize the incrementing of download counts:

* lazily load package counts and member counts for channels
* add index to package_versions
* add index to aggregated_metrics
* replace select/insert or update with upserts for postgres and sqlite


wolfv:
cool! 

Can we use the upsert for package versions as well? So that we don't see the error and can remove the try/catch block?

I think this code is awesome. I am a little afraid because it's also slightly complicated, and I still think that if we had some periodic background task on FastAPI we could simply insert the download counts every 5 minutes or so.

For example, using fastapi-utils (per this issue: https://github.com/tiangolo/fastapi/issues/520) we could use the repeated task helper which has pretty simple looking code:

https://github.com/dmontagu/fastapi-utils/blob/af95ff4a8195caaa9edaa3dbd5b6eeb09691d9c7/fastapi_utils/tasks.py#L15

btel:
> cool!
> 
> Can we use the upsert for package versions as well? So that we don't see the error and can remove the try/catch block?

sure, we should now use upsert wherever possible.

> I think this code is awesome. I am a little afraid because it's also slightly complicated, and I still think that if we had some periodic background task on FastAPI we could simply insert the download counts every 5 minutes or so.
> 
> For example, using fastapi-utils (per this issue: [tiangolo/fastapi#520](https://github.com/tiangolo/fastapi/issues/520)) we could use the repeated task helper which has pretty simple looking code:
> 
> https://github.com/dmontagu/fastapi-utils/blob/af95ff4a8195caaa9edaa3dbd5b6eeb09691d9c7/fastapi_utils/tasks.py#L15

It looks as an interesting direction to explore. Let's see how these optimizations perform in practice.  From my benchmarks, I get following latencies (95th percentile of the response time) with and without (the line commented out) incrementing download counts:

incr count | 100 concurrent requests | 10 concurrent requests
--- | --- | ---
yes | 128 ms | 975 ms
no | 44 ms | 514 ms

I ran 1000 requests for each case. These latencies do not include redirection/download times.

PR #329, opened 2021-01-29
Api Key Expiration

hbcarlos:


quetz/dao.py
```diff
@@ -617,8 +629,20 @@ def create_api_key(self, user_id, api_key: rest_models.BaseApiKey, key):
         else:
             user = User(id=uuid.uuid4().bytes)
             self.db.add(user)
+
+        print(key)
```
wolfv, line 633: if this is ready, then we should remove the print statements here

wolfv:
Awesome

PR #330, opened 2021-02-01
Batch download counts 

btel:


quetz/database.py
```diff
@@ -27,10 +40,22 @@ def get_engine(db_url, echo: bool = False, reuse_engine=True, **kwargs) -> Engin
         # TODO make configurable!
         if db_url.startswith('postgres'):
             engine = create_engine(
-                db_url, echo=echo, pool_size=32, max_overflow=100, **kwargs
+                db_url, echo=echo, pool_size=200, max_overflow=100, **kwargs
```
wolfv, line 43: Do we need to modify postgres conf for this?

PR #335, opened 2021-02-02
add support for Azure Blob Storage

madhur-tandon:


quetz/config.py
```diff
@@ -117,6 +117,17 @@ class Config:
             ],
             required=False,
         ),
+        ConfigSection(
+            "azure_blob",
+            [
+                ConfigEntry("account_name", str, default=""),
+                ConfigEntry("account_access_key", str, default=""),
+                ConfigEntry("conn_str", str, default=""),
```
wolfv, line 125: is this an commonly used abbreviation?

madhur-tandon: Yes, it's often in arguments of the official client libraries by Azure.

wolfv:
awesome!

PR #350, opened 2021-02-09
store metadata in db

madhur-tandon:


quetz/migrations/versions/546854af251f_add_channel_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+down_revision = 'ea6eba9a9ffc'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('channels', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('channel_metadata', sa.String(), nullable=True))
+        batch_op.alter_column(
```
btel, line 22: why is this generated? did you change the type of size and limit_size?

quetz/main.py
```diff
@@ -749,7 +749,10 @@ def patch_channel(
             )
 
     for attr_, value_ in user_attrs.items():
-        setattr(channel, attr_, value_)
+        if attr_ == "metadata":
```
btel, line 752: wouldn't it be better to merge with the exisitng metadata instead of replacing it. using something like this:

```
old_meta = json.loads(channel.metadata)
old_metadata.update(new_metadata)
```

quetz/db_models.py
```diff
@@ -156,6 +156,7 @@ class Channel(Base):
     private = Column(Boolean, default=False)
     mirror_channel_url = Column(String)
     mirror_mode = Column(String)
+    channel_metadata = Column(String)
```
btel, line 159: why not just `metadata`?

madhur-tandon: sqlalchemy doesn't allow it

quetz/main.py
madhur-tandon: yes perhaps, @wolfv what do you think?

quetz/migrations/versions/546854af251f_add_channel_metadata.py
madhur-tandon: I just added a new field, rest of stuff is auto-generated by alembic.
Ideally this was covered in a previous migration when we decided to use BigIntegers for sizes.
So maybe we can manually remove this?

quetz/main.py
wolfv: agreeing with btel!

quetz/migrations/versions/546854af251f_add_channel_metadata.py
madhur-tandon: Fixed in latest commit

quetz/main.py
madhur-tandon: Done

quetz/migrations/versions/ea6eba9a9ffc_merge_ebe550f9fbbe_and_b9886d9cadb0.py
```diff
@@ -0,0 +1,21 @@
+"""merge ebe550f9fbbe and b9886d9cadb0
+
```
wolfv, line 2: This one is completely empty?

madhur-tandon: yeah it just merges 2 migrations branching from the same root.

madhur-tandon: This is analogous to a merge commit. And is the standard way according to @btel 
Alternatively I manually edited it to preserve a linear history but that is dangerous if the branched migrations are deployed on beta.mamba.pm

wolfv: all good, thanks for the explanation!

PR #355, opened 2021-02-10
retry removing S3 channels on error

btel:
since deletions/additions may not be immediately visible on s3, we need to retry sometimes.

AWS S3 has strong consistency since recently: https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/

OpenStack Swift (the backend behind OVH S3) has eventual consistency:

https://julien.danjou.info/openstack-swift-consistency-analysis/


wolfv:
cool!

PR #358, opened 2021-02-10
create new sqlite connection in threads

### LINKED ISSUE #118: adding two versions of same package with sqlite backend causes segfaults
steps to reproduce:

```
# start quetz without  reload
quetz run test_quetz_test12  --dev --copy-conf dev_config.toml 

export QUETZ_API_KEY=...

curl -X POST "http://localhost:8000/api/channels" \
   -H  "accept: application/json" \
   -H  "Content-Type: application/json" \
   -H  "X-API-Key: ${QUETZ_API_KEY}" \
   -d '{"name":"channel12", "private":false }'
   
 quetz-client http://localhost:8000/api/channels/channel12 quetz/tests/data/test-package-0.1-0.tar.bz2 
 
 quetz-client http://localhost:8000/api/channels/channel12 quetz/tests/data/test-package-0.2-0.tar.bz2 
```

Results = segmentation fault

## Triaging

this might be related that the background tasks in fastapi run in separate threads. When we add files the indexes are updated in a background task using the function `quetz.indexing.update_indexes`: 

https://github.com/mamba-org/quetz/blob/d0b7b2dc331e672c32b9c8bd9917e30b509656c7/quetz/main.py#L728-L729

Running the function directly (in foreground) does not cause the segfault. The problem might be that we are re-using the same database session (attached to `dao` object) in several threads, that is known to cause problems with sqlite. postgres does not seem to have this problem (although reusing db session in mutliple threads still seem evil).

btel:
closes #118

### DIFF
--- a/quetz/deps.py
+++ b/quetz/deps.py
@@ -2,6 +2,8 @@
 Define common dependencies for fastapi depenendcy-injection system
 """
 
+import logging
+
 import requests
 from fastapi import BackgroundTasks, Depends, HTTPException, Request, status
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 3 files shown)


wolfv:
awesome!

PR #361, opened 2021-02-11
add proxylist for mirrors

wolfv:


quetz/main.py
```diff
@@ -1548,7 +1568,13 @@ async def serve_path(
         ).inc()
         download_counts[(channel.name, filename, platform)] += 1
 
-    if pkgstore_support_url and (path.endswith('.tar.bz2') or path.endswith('.conda')):
+    if is_package_request and channel.mirror_channel_url:
+        # if we exclude the package from syncing, redirect to original URL
+        proxylist = json.loads(channel.channel_metadata).get('proxylist')
```
btel, line 1573: maybe check first if metadata is not empty so that we avoid time to deserialize

wolfv: server_default is `{}` so ... I guess I can check if the len(str) <= 2

btel: 👍🏽

btel: but maybe the time will be in any case negligible

PR #371, opened 2021-02-17
clean up db and pkgstore + add platform filter in search

madhur-tandon:


quetz/authorization.py
```diff
@@ -285,6 +285,9 @@ def assert_set_channel_size_limit(self):
             msg="only server maintainer or owner can set channel size limit",
         )
 
+    def assert_channel_db_cleanup(self, channel_name):
+        self.assert_channel_roles(channel_name, [OWNER, MAINTAINER])
```
wolfv, line 289: I think only OWNER should be allowed here (maintainer might not be able to see the logs of the serveretc).

quetz/pkgstores.py
```diff
@@ -426,3 +454,15 @@ def get_filemetadata(self, channel: str, src: str):
         etag = infodata['etag']
 
         return (msize, mtime, etag)
+
+    def cleanup_temp_files(self, channel: str):
```
wolfv, line 458: this is a bit code duplication here. is there a way to make this a free function?

wolfv:
I went through all of the code and it looks really nice!

I am wondering if we can move the cleanup functions to their own python file (cleanup.py or something like that?).

Also, we could move the pkgstore cleaning function there.

quetz/authorization.py
madhur-tandon: Done.

quetz/pkgstores.py
madhur-tandon: we should tackle duplication in `S3` store and `Azure` store in a separate PR since this is a part of that now.

PR #377, opened 2021-02-22
remove check_status from job watch

btel:
since #364 check_status is no longer required - workers update themselves the task/job status


wolfv:
cool! can you squash some of the commits?

PR #389, opened 2021-03-29
add quetz_mamba_solve plugin

madhur-tandon:


wolfv:
do these tests work locally? Maybe you have to URL encode the url? Since it contains some illegal characters right now ('>='). Or you test with a simpler spec (just xtensor maybe?).

Also, it seems awkward to make this part of the URL in this way since it's "variable length". I think the canonical way to do this would be multiple `?spec=abc&spec=xyz` parameters at the end of the URL. Here is a discussion: https://stackoverflow.com/questions/2602043/rest-api-best-practice-how-to-accept-list-of-parameter-values-as-input

We can also accept JSON instead. That might be a nice interface, too...

Have you already given a thought to reloading the repodata (periodically?). I think we should reload repodata every 30 minutes when we find that it has changed.

plugins/quetz_mamba_solve/quetz_mamba_solve/api.py
```diff
@@ -0,0 +1,70 @@
...
+    return names_list
+
+
+@router.get(
+    "/api/mamba/solve/{channels}/{subdir}/{spec}", response_class=PlainTextResponse
+)
+def mamba_solve(channels, subdir, spec):
+    channels = parse_list(channels)
+    spec = parse_list(spec)
+    s = MambaSolver(channels, subdir)
```
wolfv, line 55: it would be cool to have a long-living solver where repodata is only loaded again when it's necessary. There is a replace_channel functionality that should work for this case.

wolfv:
I think actually now that using JSON as input would be better because that will allow us to extend to slightly more complex things later on (e.g. adding virtual packages through the API).

plugins/quetz_mamba_solve/quetz_mamba_solve/api.py
madhur-tandon: This is done through the time expiry based LRU cache.

PR #390, opened 2021-04-06
Add gitlab authenticator

beenje:
This PR adds GitLab as authenticator.

I realise I should have created an issue before to make this PR. This is something I need because we gonna try to deploy this internally.
I think this authenticator can be quite useful because it's quite common to have a self-hosted GitLab instance.

I didn't see any test for the GitHub or google authenticators, so there are none in this PR (it depends on the oauth2 one).

I tested the authentication manually both with gitlab.com and a self-hosted GitLab instance.


beenje:
Rebased on master to fix the tests

PR #393, opened 2021-04-11
Add Dockerfile and GitHub Workflow

beenje:
I know there is a Dockerfile under the docker directory, but it is for development.
This Dockerfile is intended more for production. It builds an image with both the frontend and conda environemnent using a multi-stages build to keep the final image size small.

I also added a GitHub workflow to build and push the image:

- The image `mambaorg/quetz:latest` is built and pushed for every commit to the master branch.
- When pushing a tag, the image is tagged with it (`mambaorg/quetz:<tag>`).
- For PR, the image is only built, but not pushed.

For the workflow to work, the `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets shall be defined for the repository.

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -1,0 +1,44 @@
+name: docker build and push
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '**'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_NAME: mambaorg/quetz
+    steps:
+      - name: Set Docker image for main branch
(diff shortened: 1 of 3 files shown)


wolfv:
looks great! 

btw. the commits use a different email address then your GitHub profile which means they might not get associated to your account. I think you can fix that if you want. Let me know then I'll hold off with merging this one.

beenje:
Thanks for noticing! I changed the email in those commits.
Will add my new work email to my profile as well.

beenje:
I forgot to mention you should have added `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets in the project settings before merging.

wolfv:
yep, adding them right now, thanks!

PR #395, opened 2021-04-12
Don't serve relative paths.

stevenryoung:
I noticed a security bug where Quetz will serve files outside of the quetz directory including root directories. Example(number of ".." may vary on your instance):

`curl --path-as-is http://quetzserver:8000/../../../../../../../../../../../../etc/passwd`

Returns my /etc/passwd file.

I was able to patch this on my instance by modifying quetz/frontend.py to handle the case where a relative path was given to static(), but I am not sure this is the best fix or a comprehensive solution.


wolfv:
thanks! I think it would be better to get an absolute path, and somehow figuring out if that path is below the frontend directory or not. But I am happy to merge this for now.

Issue #396, opened 2021-04-13
Frontend: only serve static files below frontend dir


wolfv:
This is currently fixed by checking for `..` but I think we can find a cleaner solution.

stevenryoung:
`os.path.commonpath()` seems like it could make a more robust solution (in the case that quetz should ever serve a file with two consecutive periods or expect a relative path but under the `frontend_dir`). Just check if the common path is equal to the `frontend_dir`, and make a decision based on that. https://docs.python.org/3/library/os.path.html#os.path.commonpath

wolfv:
Yes, indeed! That looks like a good function to use.

wolfv:
Thanks @stevenryoung for tackling this!

PR #400, opened 2021-05-03
Serve only files below frontend_dir using os.path.commonpath

stevenryoung:
This will use `os.path.commonpath` to ensure any static files being served are under `frontend_dir`. Currently we are just checking for `..` in the filename (see  #395), which should work, but this methods utilizes the `os.path` library which should provide a cleaner/more robust solution, addressing #396.


wolfv:
Hi @stevenryoung thanks! Looks great. I'll fix the tests later today.

wolfv:
finally got around to fix master, so let's see if this is passing fine now :)

wolfv:
Would you mind rebasing and squashing this PR?

PR #404, opened 2021-05-07
clean up frontend.py and move lab-specific stuff to quetz-frontend extension

wolfv:
We're moving the jupyter lab based frontend to a plugin that ships it's own backend helpers. 

This will also give additional freedom to other frontend implementers.

Here is the frontend repo: http://github.com/mamba-org/quetz-frontend

PS: @stevenryoung I am curious what frontend you were using so far? Would be happy to assist you in moving to the new one, or hearing your experiences.

### DIFF
--- a/quetz/frontend.py
+++ b/quetz/frontend.py
@@ -1,17 +1,13 @@
-import json
 import logging
 import os
 import sys
-from pathlib import Path
-import pkg_resources
-import jinja2
-from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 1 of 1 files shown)


stevenryoung:
@wolfv , I have only used the interface that is setup at `http://localhost:8000/` after running `quetz run test_quetz --copy-conf ./dev_config.toml --dev --reload`.  That is the interface that just helps you make your own API calls. I want to try out the other interfaces, but I have not had a chance yet.

PR #409, opened 2021-05-29
Fix startup and permission issues for docker-compose

atrawog:
This PR fixes a couple of issues in the current docker-compose.yml file:
* The startup commands for the web and init-db services use the right syntax and correctly create the quetz config and database at the first startup
* All database permission and database_url are set using a single postgres.env file. Allowing the database image to start up and ensuring the correct db settings across all images.
* The grafana admin user and password created at startup can be set to something more secure using the new grafana.env file.


wolfv:
Thanks, this is great!

PR #410, opened 2021-06-12
Remove async from serve_path

beenje:
As far as I can see, the serve_path function doesn't contain any async method.
When using blocking operation, the function shouldn't be defined as async.

See https://fastapi.tiangolo.com/advanced/custom-response/?h=stream#using-streamingresponse-with-file-like-objects

When trying to access the API at the same time as downloading files from quetz, I got very poor response time.
I think this is because serve_path was blocking the event loop.
With this change, I get much better results.

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -1552,7 +1552,7 @@ async def stop_sync_donwload_counts():
 
 
 @app.get("/get/{channel_name}/{path:path}")
-async def serve_path(
+def serve_path(
     path,
     channel: db_models.Channel = Depends(get_channel_allow_proxy),
     accept_encoding: Optional[str] = Header(None),
@@ -1650,13 +1650,13 @@ async def serve_path(
 
 
 @app.get("/get/{channel_name}")
-async def serve_channel_index(
+def serve_channel_index(
     channel: db_models.Channel = Depends(get_channel_allow_proxy),
     accept_encoding: Optional[str] = Header(None),
     session=Depends(get_remote_session),
     dao: Dao = Depends(get_dao),
 ):
-    return await serve_path("index.html", channel, accept_encoding, session, dao)
+    return serve_path("index.html", channel, accept_encoding, session, dao)
 
 
 frontend.register(app)


beenje:
The tests are failing but that doesn't look to be linked to my change.

`test_private_channels_create_package` was failing due to the new release of FastAPI (https://github.com/tiangolo/fastapi/releases/tag/0.65.2): json isn't assumed anymore. I fixed it.

wolfv:
Hi @beenje, first of all, thanks a ton for fixing the issues on master!

I am wondering if you also tried the `FileResponse`? 
https://fastapi.tiangolo.com/advanced/custom-response/?h=stream#fileresponse

It might simplify our code and we could keep the function `async`.

beenje:
> Actually, it says that StreamingResponse (which we're using) should be used with async? https://fastapi.tiangolo.com/advanced/custom-response/?h=stream#streamingresponse
> 
> Maybe the problem is that the `iter_chunks` function isn't marked as `async`?

Indeed, if we want the function to be async, all the code should be async.
If you look just below in the link you mentioned: https://fastapi.tiangolo.com/advanced/custom-response/?h=stream#using-streamingresponse-with-file-like-objects, it says:

> Notice that here as we are using standard open() that doesn't support async and await, we declare the path operation with normal def.

`StreamingResponse` can be used in an async function or not. FastAPI will take care of running it in the background.

If you look at the `FileResponse` implementation, it's using `aiofiles` to open the file asynchronously: https://github.com/encode/starlette/blob/15761fb48e4c56be09167cb8f9b761114593b651/starlette/responses.py#L304

At the end of `serve_path`, I think we could use `FileResponse`. At least for LocalStore. I'm not sure which files are served for others stores (non package request case)?
Remain the issue in [get_from_cache_or_download](https://github.com/mamba-org/quetz/blob/5cd8cb9b3ce15f618ca952743a7408d9bbd69ad9/quetz/tasks/mirror.py#L50). `remote_file = repository.open(target)` should be made async as well.

btel:
hey @beenje, thanks for debugging! I agree that there is an issue with using blocking code in async function. Your fix looks good to me - for "def" handlers FastAPI will take care of running them in a thread. This should fix some of the perf issues we saw in the past.

Alternatively, we could also make the iter_chunks an async generator as suggested by @wolfv (both in serve_path and get_from_cache_or_download) and keep the serve_path as async. Since we are not passing a file-object to StreamingResponse we don't have to worry about using "async open" (reading from file is handled by the async iter_chunks). Don't hesitate to ping me here or on gitter if you want to discuss it further.

beenje:
I agree but to keep serve_path async we'd need to change several things:

- `RemoteRepository` is using `requests` to download files. We could refactor to use `httpx` instead and make that async.
- To make iter_chunks async, we'd need the `read` to be async. But as far as I know, fsspec isn't... From the [doc](https://filesystem-spec.readthedocs.io/en/latest/async.html#async), only `HTTPFileSystem` is. `file` and `S3` aren't...

An alternative is to run the blocking code using [run_in_trheadpool](https://github.com/encode/starlette/blob/119c427474a45bd1297f5adcde864b2874dd8c4f/starlette/concurrency.py#L27).

wolfv:
@beenje we're fine to merge this :) Would you be available to rebase quickly?

beenje:
I just rebased!

wolfv:
Cool, it's merged! Thanks!

PR #411, opened 2021-06-15
Fix tests

beenje:
Tests are currently failing on master.

- I fixed test_private_channels_create_package (failure due to new version of FastAPI: https://github.com/tiangolo/fastapi/releases/tag/0.65.2)
- I fixed plugins/quetz_transmutation/tests/test_main.py::test_transmutation_endpoint. As we overwrite the headers, we need to explicitly set the content-type to application/json
- I fixed Mypy (missing stubs required). It's also possible to run "mypy --install-types" but it only supports
interactive mode right now. "--non-interactive" should be added to 0.910. See https://github.com/python/mypy/issues/10600

I moved Mypy to pre-commit, otherwise it's difficult to run locally. Any reason it wasn't there?


wolfv:
Thanks for tackling this!

beenje:
Any idea why test_transmutation_endpoint is failing?

wolfv:
Thanks!!!

wolfv:
great work!

PR #413, opened 2021-06-17
Add option to redirect static file download

beenje:
This is an attempt to improve performances when using LocalStore by letting nginx serve static files.

If `redirect_static_files` is set to `true`, quetz will return a redirect for:
- files in the cache (cached packages from proxy channels)
- files from local channels when using LocalStore

In this case, Nginx (or something else) is expected to take care of this redirect.

I tested by deploying Nginx with something like that:

```
    upstream quetz {
      server 127.0.0.1:8000;
    }

    server {
        listen      8080;
        add_header  Cache-Control $control;

        server_name  localhost;

        location / {
          proxy_set_header Host $http_host;
          proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
          proxy_set_header X-Forwarded-Proto $scheme;
          proxy_set_header Upgrade $http_upgrade;
          proxy_set_header Connection $connection_upgrade;
          proxy_redirect off;
          proxy_buffering off;
          proxy_pass http://quetz;
        }

        location /files/cache/ {
          # path for cache files
          alias /quetz-deployment/cache/;
        }

        location /files/channels/ {
          # path for channels
          alias /quetz-deployment/channels/;
        }
    }
```

By default, the new option is set to False and nothing should change.

There are probably things to discuss. I'm also not sure what should be done for other package stores.

### DIFF
--- a/quetz/config.py
+++ b/quetz/config.py
@@ -116,7 +116,7 @@ class Config:
         ConfigSection(
             "local_store",
             [
-                ConfigEntry("redirect_static_files", bool, default=False),
+                ConfigEntry("redirect_enabled", bool, default=False),
                 ConfigEntry("redirect_endpoint", str, default="/files"),
             ],
         ),
@@ -377,6 +377,7 @@ class Config:
             return pkgstores.LocalStore(
                 {
                     'channels_dir': 'channels',
+                    'redirect_enabled': self.local_store_redirect_enabled,
                     'redirect_endpoint': self.local_store_redirect_endpoint,
                 }
             )
--- a/quetz/config.py
+++ b/quetz/config.py
@@ -114,6 +114,13 @@ class Config:
             [ConfigEntry("secret", str), ConfigEntry("https_only", bool, default=True)],
         ),
         ConfigSection(
+            "local_store",
+            [
+                ConfigEntry("redirect_static_files", bool, default=False),
+                ConfigEntry("redirect_endpoint", str, default="/files"),
+            ],
+        ),
+        ConfigSection(
             "s3",
             [
                 ConfigEntry("access_key", str, default=""),
@@ -192,13 +199,6 @@ class Config:
                 ConfigEntry("channel_quota", int, required=False),
             ],
             required=False,
-        ),
-        ConfigSection(
-            "download",
-            [
-                ConfigEntry("redirect_static_files", bool, default=False),
-                ConfigEntry("redirect_endpoint", str, default="/files"),
-            ],
         ),
     ]
     _config_dirs = [_site_dir, _user_dir]
@@ -377,7 +377,7 @@ class Config:
             return pkgstores.LocalStore(
                 {
                     'channels_dir': 'channels',
-                    'redirect_endpoint': self.download_redirect_endpoint,
+                    'redirect_endpoint': self.local_store_redirect_endpoint,
                 }
             )
 
--- a/quetz/config.py
+++ b/quetz/config.py
@@ -193,6 +193,13 @@ class Config:
             ],
             required=False,
         ),
+        ConfigSection(
+            "download",
+            [
+                ConfigEntry("redirect_static_files", bool, default=False),
+                ConfigEntry("redirect_endpoint", str, default="/files"),
+            ],
+        ),
     ]
     _config_dirs = [_site_dir, _user_dir]
     _config_files = [os.path.join(d, _filename) for d in _config_dirs]
@@ -367,7 +374,12 @@ class Config:
                 }
             )
         else:
-            return pkgstores.LocalStore({'channels_dir': 'channels'})
+            return pkgstores.LocalStore(
+                {
+                    'channels_dir': 'channels',
+                    'redirect_endpoint': self.download_redirect_endpoint,
+                }
+            )
 
     def configured_section(self, section: str) -> bool:
         """Return if a given section has been configured.
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -1609,7 +1609,7 @@ async def serve_path(
         path += "index.html"
     package_content_iter = None
 
-    if config.download_redirect_static_files and hasattr(pkgstore, "redirect_url"):
+    if config.local_store_redirect_static_files and hasattr(pkgstore, "redirect_url"):
         url = pkgstore.redirect_url(channel.name, path)  # type: ignore
         return RedirectResponse(url=url)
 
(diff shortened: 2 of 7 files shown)


wolfv:
Great! Yeah, for other package stores we handle it similarly by redirecting to a pre-authenticated url. The authentication might be something to think about -- not sure if nginx has a mechanism for that

wolfv:
There is some documentation on how to implement this with NGINX here: https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/

Are you interested in tackling this? Would be an awesome feature for private channels, but I can understand if you're not interested in tacking this further :)

beenje:
I must say I'm not that interested in private channels :grin:
I will try to have a look but don't promise anything. Will be low on my list.

wolfv:
No worries. 

I think for the config option, we could change that to be in a `local_store` section (to match with the `s3` or `azure` sections?).

For the redirect itself, it might be good to make the code more equivalent to the one above (for S3 and Azure redirects?)

Maybe we can just overload the `url` function for LocalStore with the redirect url and re-use the following block:

    if is_package_request and pkgstore_support_url:
        # we have to ignore type checking here right now, sorry
        return RedirectResponse(pkgstore.url(channel.name, path))  # type: ignore

wolfv:
I will take care of private channels once this is merged, no worries about this.

beenje:
> No worries.
> 
> I think for the config option, we could change that to be in a `local_store` section (to match with the `s3` or `azure` sections?).
> 
> For the redirect itself, it might be good to make the code more equivalent to the one above (for S3 and Azure redirects?)
> 
> Maybe we can just overload the `url` function for LocalStore with the redirect url and re-use the following block:
> 
> ```
> if is_package_request and pkgstore_support_url:
>     # we have to ignore type checking here right now, sorry
>     return RedirectResponse(pkgstore.url(channel.name, path))  # type: ignore
> ```

Sounds like a good idea. I'll have to check what the url of the LocalStore was used for.

I also want to redirect for non package request (repodata.json...). I didn't really get how those files are handled for S3 and Azure?

wolfv:
For all files (including repodata.json) we redirect to the file with S3 and Azure.

quetz/tasks/mirror.py
```diff
@@ -168,6 +136,46 @@ def __getitem__(self, path):
...
+
+        return StreamingResponse(data_iter(data_stream))
+
+    if target not in cache:
+        # copy from repository to cache
+        remote_file = repository.open(target)
+        data_stream = remote_file.file
+        cache.dump(target, data_stream)
+
+    if config.local_store_redirect_enabled:
```
beenje, line 172: The `get_from_cache_or_download` function is called whatever the pkgstore. It only depends on the channel kind (proxy). Maybe using `config.local_store` isn't a good idea?

Should we use a different name?

quetz/main.py
```diff
@@ -1595,9 +1591,8 @@ async def serve_path(
         if channel_proxylist and package_name and package_name in channel_proxylist:
             return RedirectResponse(f"{channel.mirror_channel_url}/{path}")
 
-    if is_package_request and pkgstore_support_url:
-        # we have to ignore type checking here right now, sorry
-        return RedirectResponse(pkgstore.url(channel.name, path))  # type: ignore
+    if is_package_request and pkgstore.support_redirect:
```
beenje, line 1594: This is cleaner than my previous implementation but it only applies to package files (`.tar.bz2` and `.conda`).

As far as can see, json files (repodata.json) are streamed. There is no redirect. That's what the function does below. Or?

For LocalStore, I think it would make sense to use RedirectResponse as well.

quetz/config.py
```diff
@@ -113,6 +113,13 @@ class Config:
             "session",
             [ConfigEntry("secret", str), ConfigEntry("https_only", bool, default=True)],
         ),
+        ConfigSection(
+            "local_store",
+            [
+                ConfigEntry("redirect_enabled", bool, default=False),
+                ConfigEntry("redirect_endpoint", str, default="/files"),
```
beenje, line 120: We could maybe only define one option: if redirect_endpoint is set to an empty string, redirect would be disabled.
But I think it's a bit clearer like that.

beenje:
Note than when we agreed on the changes, I'll try to update the documentation with an example configuration for Nginx.

quetz/tasks/mirror.py
wolfv: hmm, we should rather use the pkgstore interface here, shouldn't we?
I need to check what type LocalCache is.

wolfv: OK, I just checked and looks like the localcache implementation is a bit disjoint from our pkgstore interface that would otherwise be easily usable to generate URLs. 

I think it would be worthwhile to rewrite this part of the code to use the pkgstore interface (to be able to proxy-cache on S3 / Azure as well). In that sense, I still think it makes sense to use the variables from the local_store and say that we should rewrite the localcache to support the pkgstore.

quetz/main.py
wolfv: Yeah, the reasons for that is that we need to add the headers for cache control, and we try to do some more intelligent stuff around gzip encoding of these JSON files.
The JSON files without GZIP encoding are really large (>100Mb vs ~15Mb).

quetz/tasks/mirror.py
beenje: I just pushed some changes to remove the `LocalCache`. I copy the files to the package store instead.
I did some quick tests and it seems fine. I don't think there is an issue with the conda packages.
For the repodata.json and current_repodata.json, I saw those files are created (empty) when creating the proxy channel. There are also a repodata.json.gz that won't be updated (which could be an issue I guess).

quetz/main.py
```diff
@@ -1559,14 +1556,20 @@ async def serve_path(
...
     if channel.mirror_channel_url and channel.mirror_mode == "proxy":
         repository = RemoteRepository(channel.mirror_channel_url, session)
-        return get_from_cache_or_download(repository, cache, path)
+        if not pkgstore.file_exists(channel.name, path):
+            download_remote_file(repository, pkgstore, channel.name, path)
+        elif path.endswith(".json"):
+            # repodata.json and current_repodata.json are cached locally
+            # for channel.ttl seconds
+            _, fmtime, _ = pkgstore.get_filemetadata(channel.name, path)
+            if time.time() - fmtime >= channel.ttl:
```
beenje, line 1571: I think using the channel ttl makes sense here but I noticed I couldn't change the default value when defining the proxy channel.
The default is 10 hours which is probably too long for this use case. Is that on purpose?

wolfv:
wow, this is great :)

PR #415, opened 2021-06-21
Allow to change the channel ttl

beenje:
This PR allows the user to change the channel ttl.

The default value of 10 hours sounds a bit high to me, but it's not a big issue if it can be changed.


wolfv:
Thanks!

Issue #414, opened 2021-06-18
Try to cache repodata.json for proxy channels


beenje:
Currently the repodata.json and current_repodata.json files aren't cached for proxy channels: https://github.com/mamba-org/quetz/blob/bfe0af1c3e668de0948b2b3cb093cd451d9a6493/quetz/tasks/mirror.py#L67

The file is streamed every time a client requests it. We could cache it locally to avoid downloading it for every clients.
Anaconda sets max-age to 1200 seconds in the cache-control header.

Maybe something similar is already done for mirror channels?

wolfv:
Yes, we do also set cache control headers. 
I think it might be even better to simply redirect? Since that would set all headers properly. Otherwise we should forward headers and make sure that we get a gzip-encoded response.

beenje:
I'd like to redirect, but that needs to be a local file. We can't redirect to the original url as the point of using a proxy channel is probably that clients are behind a firewall and can't access the original channel.

beenje:
This was implemented in https://github.com/mamba-org/quetz/pull/413
Closing.

PR #416, opened 2021-06-30
Add documentation about Nginx for static file redirect

beenje:
- Serve all files with Nginx when using LocalStore (including `repodata.json`)
- Add nginx config to docker-compose for local testing
- Add documentation about nginx configuration

I think serving json files with Nginx when it is already configured for channels makes sense as it can compress json files and add max-age to the cache-control header (as shown in the example and documentation). The only drawback I see is that the `max-age` will be the same for all channels. It has to be hard-coded in Nginx configuration and can't be set to the channel ttl.

This is only enabled for the LocalStore as Nginx needs to be configured. Maybe it should be enabled via a different option so it can be used when using another package store?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ version: '3'
 services:
   web:
     image: quetz_dev_image
+    container_name: quetz-web
     ports:
       - "8000:8000"
     build: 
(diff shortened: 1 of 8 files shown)


beenje:
This would fix #414

PR #426, opened 2021-07-22
add redirect to nginx with secure link

wolfv:
@beenje would be interested in your feedback here.


beenje:
From a quick look, this looks good to me!
Note that I haven't tested it.

Tests will have to be fixed (due to new settings) :-)

Issue #429, opened 2021-08-11
Session may be inconsistent with database


adriendelsalle:
Description
---

The session may be inconsistent with the database when re-using the same secret.
For example it happens when you run a new server using `--delete` option but still re-use the same config (the user id from the session doesn't exist anymore in the database).
It may also happen when deleting your own user, the session is still active and you can continue to act as this user.

We could probably:
- make a database request to ensure the user exists in `assert_user` or `get_user`
  - I would say `assert_user`, `get_user` doesn't explicitly makes me think there is a check
- replace the session secret with some randomly generated value, even when using the `--copy-conf` flag

PR #430, opened 2021-08-11
Make `assert_user` check that user exists in db

adriendelsalle:
Description
---

The session may contain a `user_id` that is not in the database anymore.
Make `assert_user` catch this case

Related to https://github.com/mamba-org/quetz/issues/429

### DIFF
--- a/quetz/authorization.py
+++ b/quetz/authorization.py
@@ -62,7 +62,7 @@ class Rules:
     def assert_user(self) -> bytes:
         user_id = self.get_user()
 
-        if not user_id:
+        if not user_id or not self.db.query(User).filter(User.id == user_id).count():
             raise HTTPException(
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail="Not logged in",


wolfv:
lgtm 

restarted the test to see if it's just flaky

wolfv:
cc @beenje just FYI it seemed like one of the concurrent fetch tests is slightly flaky. Specifically in this test https://github.com/mamba-org/quetz/blob/7485a98d2f678f3993c222f2639337d5dd5679dd/quetz/tests/test_mirror.py#L635-L636 the lefthand-side contained two entries.

quetz/authorization.py
```diff
@@ -62,7 +62,7 @@ def get_user(self) -> Optional[bytes]:
     def assert_user(self) -> bytes:
         user_id = self.get_user()
 
-        if not user_id:
+        if not user_id or not self.db.query(User).filter(User.id == user_id).count():
```
wolfv, line 65: shoot, we should have called logout(session) when this check fails.

beenje:
> cc @beenje just FYI it seemed like one of the concurrent fetch tests is slightly flaky. Specifically in this test
> 
> https://github.com/mamba-org/quetz/blob/7485a98d2f678f3993c222f2639337d5dd5679dd/quetz/tests/test_mirror.py#L635-L636
> 
> the lefthand-side contained two entries.

Thanks for letting me know. I'll try to reproduce and see if I can improve things.

PR #433, opened 2021-08-25
add debug info and use bytes for user_id

atrawog:
Fixes the "reindex_packages_from_store:can't escape str to binary" error when trying to reindex an already existing S3 bucket.


wolfv:
in the environment.yml can you change from `repo.mamba.pm` to `conda-forge`? Until we have the server back up, at least :)

PR #438, opened 2021-08-31
add IDEs to gitignore

adriendelsalle:
Description
---

Add IDEs (vscode and pycharm) files to gitignore file


JohanMabille:
This is considered as bad practice and should be done in your local git configuration.

adriendelsalle:
Thanks for the pointer!

PR #439, opened 2021-08-31
Factorize `get_db_manager`

adriendelsalle:
Description
---

add `get_db_manager` to database module
reuse `database.get_db_manager` in plugins
update  `get_session` mock patch to take any args or kwargs
add docstring to mention that `get_session` function is mocked during tests
side modifications:
- add sql echo fixture
- add some type hints

### DIFF
--- a/quetz/testing/fixtures.py
+++ b/quetz/testing/fixtures.py
@@ -40,8 +40,14 @@ def database_url(sqlite_url):
 
 
 @fixture
-def engine(database_url):
-    sql_echo = bool(os.environ.get("QUETZ_TEST_ECHO_SQL", False))
+def sql_echo():
+    """whether to activate SQL echo during the tests or not."""
+    return False
+
+
+@fixture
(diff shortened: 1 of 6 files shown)


quetz/testing/fixtures.py
```diff
@@ -40,8 +40,14 @@ def database_url(sqlite_url):
 
 
 @fixture
-def engine(database_url):
-    sql_echo = bool(os.environ.get("QUETZ_TEST_ECHO_SQL", False))
+def sql_echo():
+    """whether to activate SQL echo during the tests or not."""
+    return False
```
wolfv, line 45: Is this on purpose? Why not use the env var here?

adriendelsalle: it allows to override it using this fixture, see the modification in `engine`: it defaults sql_echo to the env var value

wolfv: I find it quite confusing tbh. Do we really need it?

PR #440, opened 2021-09-01
Compatibility with `alembic` 1.7

adriendelsalle:
Description
---

fix tests using `alembic` 1.7
use `importlib` to load `entry_points`
remove use of `pkg_resources` to be consistent with `alembic` internal implementation
rename `dummy_plugin` fixture to `dummy_job_plugin` to be more explicit
rename `entry_points` fixture to `dummy_migration_plugin`
refactor `dummy_migration_plugin` to mock a package and its `entry_points`

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -13,11 +13,11 @@ from multiprocessing import get_context
 from pathlib import Path
 from typing import NoReturn, Optional, Union
 
-import pkg_resources
 import typer
 import uvicorn
 from alembic import command
 from alembic.config import Config as AlembicConfig
+from importlib_metadata import entry_points
 from sqlalchemy.orm.session import Session
 from sqlalchemy_utils.functions import database_exists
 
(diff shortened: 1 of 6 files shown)


adriendelsalle:
Don't know why coverage decreases, nothing changed about `pkgstores` code or tests

quetz/cli.py
```diff
@@ -132,13 +132,13 @@ def _make_migrations(
     if plugin_name == "quetz":
         version_path = None  # Path(quetz.__file__).parent / 'migrations' / 'versions'
     else:
-        entry_point = next(
-            pkg_resources.iter_entry_points('quetz.migrations', plugin_name)
-        )
+        entry_point = tuple(
```
wolfv, line 135: does the `next` trick not work anymore?

wolfv:
Looks good!

quetz/cli.py
adriendelsalle: It is but I think it would raise a deprecation warning. It's recommended to cast to tuple instead

wolfv: Who recommends that? It could hypothetically use a lot more space etc

PR #443, opened 2021-09-02
use multiple threads for re-indexing packages

atrawog:
This PR makes re-indexing packages from store multi-threaded to speed up loading package info from the package store and implements a sync option to only update packages that aren't already present in the database.

### DIFF
--- a/quetz/tasks/reindexing.py
+++ b/quetz/tasks/reindexing.py
@@ -1,7 +1,9 @@
 import json
 import logging
 import os
+import time
 import uuid
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from sqlalchemy.exc import IntegrityError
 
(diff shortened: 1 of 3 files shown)


quetz/tasks/reindexing.py
```diff
@@ -43,57 +70,116 @@ def handle_file(channel_name, filename, file_buffer, dao, user_id):
...
+
+        toc = time.perf_counter()
+        logger.debug(
+            f"Imported files {pkg_group[0]} to {pkg_group[-1]} "
+            + f"for channel {channel_name} in {toc - tic:0.4f} seconds "
+            + f"using {nthreads} threads"
+        )
+
+        try:
+            update_indexes(dao, pkgstore, channel_name)
```
wolfv, line 179: this happens every 128 packages now, correct?

I am just asking because there is one step in `update_indexes` that takes a bit of time (generating the current_repodata.json necessary for `conda`).

PR #444, opened 2021-09-03
make package_queue_size configurabe

atrawog:
This PR imports all versions of a package at the same time (instead of first importing packages version sorted by architecture) and adds an option package_queue_size to make it configurable how many packages will be imported before a new repodata.json file is generated.

### DIFF
--- a/quetz/config.py
+++ b/quetz/config.py
@@ -71,6 +71,7 @@ class Config:
             "general",
             [
                 ConfigEntry("package_unpack_threads", int, 1),
+                ConfigEntry("package_queue_size", int, 1),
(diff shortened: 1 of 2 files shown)


quetz/config.py
```diff
@@ -71,6 +71,7 @@ class Config:
             "general",
             [
                 ConfigEntry("package_unpack_threads", int, 1),
+                ConfigEntry("package_queue_size", int, 1),
```
wolfv, line 74: hmmm ... this is pretty specific to the reindexing task ... couldn't this be an arguement to the task, actually?

atrawog: Yes, I can turn it into an argument for the task.

PR #449, opened 2021-09-06
feat: implement google cloud storage package store

baszalmstra:
Initial implementation of a Google Cloud Storage backend. Although Google Cloud Storage can also be used through the S3 backend this resulted in some errors. This PR adds an implementation based on `gcsfs` which uses the Google Cloud Storage API directly.

Note that all package data is served through the Quetz server which will result in higher performance requirements. This is disabled for the GCS store because signing of URLs is currently not yet supported by `gcsfs`.


quetz/config.py
```diff
@@ -146,6 +146,17 @@ class Config:
             ],
             required=False,
         ),
+        ConfigSection(
+            "gcs",
+            [
+                ConfigEntry("project", str, default=""),
+                ConfigEntry("token", str, default=""),
+                ConfigEntry("conn_str", str, default=""),
```
wolfv, line 154: is `conn_str` used anywhere?

wolfv:
Besides that one comment, this looks good to me. I am happy to merge and we can add the `url` functionality later on, when integrated upstream.

quetz/config.py
baszalmstra: Good catch! I removed it!

wolfv:
Thanks @baszalmstra !

PR #452, opened 2021-09-08
fix: gcs bucket exists

baszalmstra:
If a bucket already exists `mkdir` in gscfs raises an HTTPError: "You already own this bucket.Please select another name., 409". I modified the behavior of `create_channel` to check if the bucket exists before creating it. This will usually result in the same number of API calls since most of the time the bucket will exist.

### DIFF
--- a/quetz/pkgstores.py
+++ b/quetz/pkgstores.py
@@ -595,10 +595,10 @@ class GoogleCloudStorageStore(PackageStore):
             The name of the container to create on azure blob
         """
         with self._get_fs() as fs:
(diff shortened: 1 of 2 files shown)


baszalmstra:
I've added a test to verify this behavior. 

We have to invalidate the filesystem cache after creating a bucket because `gcsfs`' cache is to aggressive. See [this issue in `gcsfs`](https://github.com/dask/gcsfs/issues/413).

wolfv:
Looks great! Thanks

PR #455, opened 2021-09-16
upload repodata to S3 with put_file

atrawog:
Passing large repodata files in memory is inefficient and can fail for large repodata files like conda-forge.
This PR instead uses the s3fs.put_file() to transfer the generated repodata files directly to S3.


wolfv:
Looks great! Happy to merge. 

Can you re-open the PR from a different branch? Then we don't have these "merge branch ... into master" commits.

Best practice would be to create a new branch (different from master) in your fork for every PR.

wolfv:
OK I just squashed it, so that's fine. But in general PR's from other branches than master are cleaner :)

PR #463, opened 2021-10-13
terms of service plugin

madhur-tandon:


plugins/quetz_tos/quetz_tos/api.py
```diff
@@ -0,0 +1,118 @@
...
+        if not selected_tos:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"terms of service with id {tos_id} not found",
+            )
+    else:
+        selected_tos = db.query(TermsOfService).order_by(TermsOfService.time_created.desc()).first()
+
+    if selected_tos:
+        signature = db.query(TermsOfServiceSignatures).filter(TermsOfServiceSignatures.user_id == user_id).filter(TermsOfServiceSignatures.tos_id == selected_tos.id).one_or_none()
```
wolfv, line 94: I think you can merge the filter when using an _and

PR #464, opened 2021-10-18
Use FPS to register routers

davidbrochart:


setup.cfg
```diff
@@ -55,8 +53,10 @@ install_requires =
 [options.entry_points]
 console_scripts =
     quetz = quetz.cli:app
+fps_router =
+    quetz = quetz.main
```
davidbrochart, line 57: FPS doesn't seem to take into account the routers defined in `quetz/main.py`. Any idea why @adriendelsalle ?

adriendelsalle: looks like you use the `@app` decorator instead of `@api_router`?

PR #470, opened 2021-10-31
Fix bad package references

maresb:
Both mamba and quetz from PyPI are unrelated!

* [mamba: the definitive test runner for Python](https://pypi.org/project/mamba/)
* [quetz: A Python library for compositional programming.](https://pypi.org/project/quetz/)

### DIFF
--- a/plugins/quetz_mamba_solve/requirements.txt
+++ b/plugins/quetz_mamba_solve/requirements.txt
@@ -1,2 +1,1 @@
-mamba
 conda-build >=3.20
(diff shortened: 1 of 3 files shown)


plugins/quetz_mamba_solve/requirements.txt
```diff
@@ -1,2 +1 @@
-mamba
```
wolfv, line 1: hmm, I think we could keep this here since it's used when doing `quetz plugin install quetz_mamba_solve`.

Although we could maybe refactor this across the board and use a different filename.

maresb:
Starting over in #471

PR #474, opened 2021-11-02
Fix compatibility with SQLAlchemy 1.4

adriendelsalle:
Description
---

Fix compatibility with SQLAlchemy>=1.4:
- use `LargeBinary` instead of `Binary`
- also change existing `alembic` migration since `LargeBinary` is a synonym of `Binary` (deprecated since 2010 and removed in `sqlalchemy 1.4.0`, see https://github.com/sqlalchemy/sqlalchemy/issues/6263#issuecomment-819645247)
- remove upper bounds on `conda` and `pip` package deps

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - authlib
   - psycopg2
   - httpx=0.20.0
-  - sqlalchemy<1.4.0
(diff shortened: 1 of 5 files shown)


wolfv:
Nice!

PR #478, opened 2022-01-10
Add releaser

fcollonval:
This add the jupyter releaser to publish Quetz python package as _quetz-server_ on PyPi

The dependencies were splitted into extras to reflect their optional character.


fcollonval:
Kicking the CI

wolfv:
if you rebase on master the mamba solve problems should be fixed now.

fcollonval:
This is ready for review - I don't know enough the plugins to figure out why some are failing

wolfv:
I am not sure if we should really expand the matrix like this. Could we keep the previous way of testing plugins in another step of a given run?

fcollonval:
> I am not sure if we should really expand the matrix like this. Could we keep the previous way of testing plugins in another step of a given run?

The idea to split the test between plugins is to spot directly which one fails. Otherwise, if multiple are failing, you need multiple runs to know.

.github/workflows/ci.yml
```diff
@@ -111,14 +97,23 @@ jobs:
...
           export PATH=$(dirname $MAMBA_EXE):$PATH
           echo "adding micromamba to path: $MAMBA_EXE"
 
           pip install --no-deps git+https://git@github.com/regro/libcflib@master
           
-          quetz plugin install ./plugins/${{ matrix.plugin}}
-          pytest -v ./plugins/${{ matrix.plugin }}
+          for f in ./plugins/quetz_*
+          do
+            echo "Testing plugin" $f
```
wolfv, line 113: I think pytest might do some of this automatically, but there are fancy ways of formatting in github actoins with e.g. we could group log lines with this https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

E.g. using `echo ::group::$f` and then `echo ::endgroup::` at the end

fcollonval: Thanks for the suggestion

wolfv:
Awesome! Thanks!

PR #495, opened 2022-03-08
Allow to delete channel members and change role of existing members

### LINKED ISSUE #494: Allow deleting channel members
Currently, there is no way to delete a channel member or change its role, [as the server raises a 409 error if the user is already a channel member](https://github.com/mamba-org/quetz/blob/4d88b4695166d310823a48e81e025983846afd05/quetz/main.py#L914-L931).

janjagusch:
Closes #494

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -916,12 +916,19 @@ def post_channel_member(
     new_member: rest_models.PostMember,
     channel: db_models.Channel = Depends(get_channel_or_fail),
     dao: Dao = Depends(get_dao),
+    db=Depends(get_db),
     auth: authorization.Rules = Depends(get_rules),
 ):
 
     auth.assert_add_channel_member(channel.name, new_member.role)
 
-    dao.create_channel_member(channel.name, new_member)
+    channel_member = dao.get_channel_member(channel.name, new_member.username)
+
+    if channel_member:
+        channel_member.role = new_member.role
+        db.commit()
+    else:
+        dao.create_channel_member(channel.name, new_member)
 
 
 @api_router.delete("/channels/{channel_name}/members", tags=["channels"])
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -921,13 +921,6 @@ def post_channel_member(
 
     auth.assert_add_channel_member(channel.name, new_member.role)
 
-    channel_member = dao.get_channel_member(channel.name, new_member.username)
-    if channel_member:
-        raise HTTPException(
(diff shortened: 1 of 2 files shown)


quetz/main.py
```diff
@@ -908,27 +907,64 @@ def get_channel_members(
...
     member_list = dao.get_channel_members(channel.name)
 
-    return member_list
+    return [
+        rest_models.PostMember(
+            username=member.user.username,
+            role=member.role,
+        )
+        for member in member_list
+    ]
```
janjagusch, line 916: I'm not really happy with having to change the return type here. Again, the main issue here is that `post_channel_members` and  `get_channel_members` disagree on whether the role can be `None`.

An alternative would be to allow `role=None` in the `rest_models.Member` class. It would also be possible to completely ban `role=None` for channel members, but since there is a difference between being a channel member with `role=None` and not being a channel member, I would advise against that.

quetz/main.py
```diff
@@ -916,19 +916,50 @@ def post_channel_member(
...
+
+    auth.assert_list_channel_members(channel.name)
+    channel_member = dao.get_channel_member(channel.name, username)
+
+    if not channel_member:
         raise HTTPException(
-            status_code=status.HTTP_409_CONFLICT,
-            detail=f"Member {new_member.username} in {channel.name} exists",
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"channel member {username}/{channel.name} not found",
```
wolfv, line 956: Maybe it would be clearer to write "User {username} does not exist or is not a member of {channel.name}" or something along those lines?

wolfv:
This looks good to me. 

Let me know what you think about my comment.

Are you OK if I squash the commits when merging? Or do you want to clean up the history?

quetz/main.py
janjagusch: Sure! I'll go ahead and adjust it.

janjagusch: should be good now

janjagusch:
> This looks good to me.

Great!

> Let me know what you think about my comment.

I've incorporated your comment.

> Are you OK if I squash the commits when merging? Or do you want to clean up the history?

Totally fine with squashing. Sorry about the commit mess, I think I was debugging against CI at some point.

janjagusch:
> I'll merge after CI passes!

Nice! Apparently, there are some more CI issues, which also seem to occur on `master`: https://github.com/mamba-org/quetz/runs/5508578696?check_suite_focus=true

Should be unrelated to this PR then, right?

PR #506, opened 2022-03-15
Serve repodata.json with redirect

jonashaag:
Using a redirect is faster than streaming it through the Quetz instance.

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -1651,37 +1651,38 @@ def serve_path(
             download_remote_file(repository, pkgstore, channel.name, path)
             fsize = fmtime = fetag = None
 
-    gzip_exists = (
-        is_repodata_request
-        and accept_encoding
-        and 'gzip' in accept_encoding
-        and pkgstore.file_exists(channel.name, path + ".gz")
-    )
-
     # Redirect response
     if (is_package_request or is_repodata_request) and pkgstore.support_redirect:
-        return RedirectResponse(
-            pkgstore.url(channel.name, path + ".gz" if gzip_exists else path)
-        )
+        return RedirectResponse(pkgstore.url(channel.name, path))
 
     # Streaming response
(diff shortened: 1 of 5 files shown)


quetz/main.py
```diff
@@ -1630,65 +1628,74 @@ def serve_path(
...
+    gzip_exists = (
+        is_repodata_request
+        and accept_encoding
+        and 'gzip' in accept_encoding
+        and pkgstore.file_exists(channel.name, path + ".gz")
+    )
 
+    # Redirect response
+    if (is_package_request or is_repodata_request) and pkgstore.support_redirect:
+        return RedirectResponse(
```
wolfv, line 1663: did you confirm that this works? Since it redirects to a path with `.gz` attached and I am not sure mamba / conda will handle that correctly.

jonashaag:
@wolfv works now. We still get a 403 instead of a 404 but I guess that's difficult to solve without adding a file-exists check before the redirect (which would be costly)

quetz/pkgstores.py
```diff
@@ -376,7 +380,11 @@ def remove_prefix(text, prefix):
         with self._get_fs() as fs:
             return [remove_prefix(f, channel_bucket) for f in fs.find(channel_bucket)]
 
-    def url(self, channel: str, src: str, expires=3600):
+    def url(
+        self, channel: str, src: str, expires=3600, content_type: Optional[str] = None
+    ):
+        if content_type is not None:
+            raise NotImplementedError("'content_type' is not supported for S3")
```
jonashaag, line 387: TODO: Check what to do with S3. I am not sure if S3 will happily accept non-signed headers or will error out (in the latter case we can't use redirects for `repodata.json` on S3)

janjagusch (commented):
I'm no longer maintaining this repository.

PR #507, opened 2022-03-16
Move httpx as dependency

fcollonval:
httpx is actually an indirect dependency required by the authentication package

This was seen when testing the package for quetz-frontend


wolfv:
Hm, tests are failing...

fcollonval:
Blaming the authlib 1.0.0 release for this one: https://github.com/lepture/authlib/releases

PR #526, opened 2022-05-02
Fix PAM authentication log message

riccardoporreca:
Missing f-string and spacing


riccardoporreca:
Happy to also convert the `%` formatting of other log messages to `f` strings for consistency

wolfv:
@riccardoporreca if you want to tackle other strings to convert them to `f"...` that would be great. 
Thansk for this PR!

PR #529, opened 2022-05-11
fix proxy channels noarch and gzip repodata

wolfv:
@janjagusch this fixes the `noarch` issue for proxy channels and also adds the same `gzip` mechanism we use with regualr channels which should speed up repodata retrieval substantially (if you're not encoding the stream elsewhere anyways).

Let me know what you think.


wolfv:
@janjagusch looks like the CI issue is unrelated. Looking into that now.

wolfv:
This should be green now.

PR #532, opened 2022-05-12
Log post_index_creation exceptions

atrawog:
Quetz allows to add plugins to modify the repodata info before they are written to disk or bucket.

This PR adds proper exceptions handling to log plugin exceptions and prevent cases where a plugin error causes no repodata be written to disk at all.

### DIFF
--- a/quetz/tasks/indexing.py
+++ b/quetz/tasks/indexing.py
@@ -144,10 +144,7 @@ def validate_packages(dao, pkgstore, channel_name):
                         f"File size differs for {filename}: "
                         f"{f['size']} vs {db_dict[filename]}"
                     )
-                    pkgstore.delete_file(
-                        channel_name,
-                        f"{subdir}/{filename}",
-                    )
+                    pkgstore.delete_file(channel_name, f"{subdir}/{filename}")
                     db_pkg_to_delete = (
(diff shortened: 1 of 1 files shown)


quetz/tasks/indexing.py
```diff
@@ -210,11 +207,14 @@ def update_indexes(dao, pkgstore, channel_name, subdirs=None):
...
-            subdir=sdir,
-        )
+        try:
+            logger.debug(f"Starting post_index_creation for {sdir} of {channel_name}")
+            pm.hook.post_index_creation(
+                raw_repodata=raw_repodata, channel_name=channel_name, subdir=sdir
+            )
+            logger.debug("Finished post_index_creation for {sdir} of {channel_name}")
+        except Exception:
+            logger.exception("Exception post_index_creation:")
```
wolfv, line 217: Is there something missing? exception isn't part of the logger? or is there some magic when usign `logger.exception`?

wolfv: Ah, OK, I looked it up, and there is some magic with `logger.exception` that I wasn't aware of :)

atrawog: Yes ```logging.exception``` provides some syntactic sugar to make exception logging a bit easier.

PR #533, opened 2022-05-16
Add upload endpoint

atrawog:
This PR adds the api endpoint  ```/channels/{channel_name}/upload/{filename}``` to quetz. The new endpoint includes a mandatory sha256 checksum verification and uses an async ```request.stream``` that is calculating the sha256 checksum while uploading.

An example file upload for  [sqlite-3.37.2-hc67169a_0.tar.bz2](https://beta.mamba.pm/get/emscripten-forge/emscripten-32/sqlite-3.37.2-hc67169a_0.tar.bz2) with curl using the new endpoint looks like this:
```
curl -i -H  "X-API-Key: ${QUETZ_API_KEY}" \
-X POST http://127.0.0.1:8000/api/channels/uptest/upload/sqlite-3.37.2-hc67169a_0.tar.bz2?sha256=63da44f64fb627ad144d38df48b24fb741f8f787027b38219273b40ac75bada9 \
--data-binary @sqlite-3.37.2-hc67169a_0.tar.bz2
```

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -2,6 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 import asyncio
 import datetime
+import hashlib
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


wolfv:
This looks good to me. The only modification I would like is to also expect either a `Content-Length` header or a `?size=...` argument to have knowledge about the expected file size. While this doesn't give any additional guarantees vs. the hash, I think it could be quite useful to later e.g. abort an upload early if not enough space is available and so on.

PR #538, opened 2022-05-30
Fix / consistent usage of f-strings

riccardoporreca:
Follow-up from #526

* Fix missing `f` for massage strings with placeholders otherwise unresolved.
* Favor f-strings over `.format()` when sensible to do so.
* Replace %-formatting with f-strings.

### DIFF
--- a/quetz/jobs/runner.py
+++ b/quetz/jobs/runner.py
@@ -50,7 +50,7 @@ def sqlite_all(element, compiler, **kw):
 
 @compiles(all_true, 'postgresql')
 def pg_all(element, compiler, **kw):
-    return f'bool_and({compiler.process(element.clauses, **kw)}'
+    return f'bool_and({compiler.process(element.clauses, **kw)})'
 
 
 def build_queue(job):
--- a/quetz/jobs/runner.py
+++ b/quetz/jobs/runner.py
@@ -29,12 +29,12 @@ class any_true(FunctionElement):
 
 @compiles(any_true, 'sqlite')
 def sqlite_any(element, compiler, **kw):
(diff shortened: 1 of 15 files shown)


quetz/jobs/runner.py
```diff
@@ -45,12 +45,12 @@ class all_true(FunctionElement):
...
 @compiles(all_true, 'sqlite')
 def sqlite_all(element, compiler, **kw):
-    return 'min(%s)' % compiler.process(element.clauses, **kw)
+    return f'min({compiler.process(element.clauses, **kw)})'
 
 
 @compiles(all_true, 'postgresql')
 def pg_all(element, compiler, **kw):
-    return 'bool_and(%s)' % compiler.process(element.clauses, **kw)
+    return f'bool_and({compiler.process(element.clauses, **kw)}'
```
baszalmstra, line 53:
```suggestion
    return f'bool_and({compiler.process(element.clauses, **kw)})'
```

baszalmstra:
You could also consider running: https://github.com/asottile/pyupgrade

But that won't catch the many errors you fixed as well! :)

riccardoporreca:
I have actually found and used a flake8 plugin https://pypi.org/project/flake8-use-fstring/ to easily spot `%` and `.format()`.

riccardoporreca:
@wolfv, @baszalmstra, I am coming back to this: I have included the `flake8-use-fstring` plugin and changed an additional occurrence of `.format`.
I have checked whether the `main` branch might have introduced meanwhile similar issue with unresolved placeholders (not detectable by the plugin) and could not find any such instance.

PR #541, opened 2022-06-21
channels with dots in their name cause a crash in indexing

### LINKED ISSUE #540: Channel name with dot causes crash
Currently, when creating a channel with a dot inside (`my.channel`), then there is a crash in the indexing task.

```py
File "/env/lib/python3.8/site-packages/quetz/main.py", line 726, in post_channel
indexing.update_indexes(dao, pkgstore, new_channel.name)
File "/env/lib/python3.8/site-packages/quetz/tasks/indexing.py", line 256, in update_indexes
to_upload = open(path, 'rb')
IsADirectoryError: [Errno 21] Is a directory: '/tmp/tmp7w_5j_c7/my.channel'
```

The cause is that the indexing globs `*.*` and then catches the directory as well: https://github.com/mamba-org/quetz/blob/d63e2f155aa75dfe852b70674b01ab7732d234c2/quetz/tasks/indexing.py#L254

I am unsure about the fix.. limiting the iteration to the channel itself seems to be over-limiting, maybe the task is being used on a collection of channels at once? if not we could just skip non-files?

gabm:
Solution: skip non-files

fixes #540

### DIFF
--- a/quetz/tasks/indexing.py
+++ b/quetz/tasks/indexing.py
@@ -252,6 +252,9 @@ def update_indexes(dao, pkgstore, channel_name, subdirs=None):
     tmp_suffix = uuid.uuid4().hex
     after_upload_move = []
     for path in tempdir_path.rglob('*.*'):
+        # check whether the path is an actual file. this fixes https://github.com/mamba-org/quetz/issues/540
+        if not path.is_file():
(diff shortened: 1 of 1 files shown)


wolfv:
Thanks!

PR #543, opened 2022-06-23
first modifications for language of TOS

MichaelKora:


martinRenou (changes requested):
Thanks!! Looks great. We probably need upload to support a `language` parameter as well?

plugins/quetz_tos/tests/test_quetz_tos.py
```diff
@@ -39,9 +39,15 @@ def test_tos_upload_by_owner(client, owner_user):
...
 def test_get_tos(client, tos_file, tos):
-    response = client.get('/api/tos')
+    params = {'language': 'EN'}
+    response = client.get('/api/tos', params=params)
+    assert response.json()['filename'] == 'tos.txt'
+    assert response.json()['content'] == 'demo tos'
+
+def test_get_tos_fr(client, tos_file, tos):
+    params = {'language': 'FR'}
+    response = client.get('/api/tos', params=params)
```
martinRenou, line 50: is it equivalent to `response = client.get('/api/tos?language=FR')`?

plugins/quetz_tos/quetz_tos/api.py
```diff
@@ -33,11 +33,15 @@ def post_file(file):
     return file.filename
 
 
-@router.get("/api/tos", tags=['Terms of Service'])
-def get_current_tos(db: Session = Depends(get_db)):
+@router.get("/api/tos/?{lang}", tags=['Terms of Service'])
```
martinRenou, line 36: I guess this does not work?

For adding query parameters you can look at how it is done here: https://github.com/tiangolo/fastapi#create-it. See how they have the `"q"` query parameter.

So you probably want something like:
```python
@router.get("/api/tos", tags=['Terms of Service'])
def get_current_tos(lang: str = "EN", db: Session = Depends(get_db)):
```

And then this API will be called with `GET /api/tos?lang=FR`

plugins/quetz_tos/quetz_tos/api.py
```diff
@@ -33,11 +33,15 @@ def post_file(file):
...
+def get_current_tos(lang: str, db: Session = Depends(get_db)):
+    #lang = request.args.get('language')
+
     current_tos = (
-        db.query(TermsOfService).order_by(TermsOfService.time_created.desc()).first()
+        #db.query(TermsOfService).order_by(TermsOfService.time_created.desc()).first()
+        db.query(TermsOfService).filter_by(language == f"{lang}").first()
```
martinRenou, line 42: Does it work? It seems `language` is not defined here?

For the code styling, `f"{lang}"` can simply be `lang`.

Maybe the following works? (fetch all terms of services and get the first one that matches the language)
```python
terms_of_services = (
    db.query(TermsOfService).order_by(TermsOfService.time_created.desc()).all()
)

current_tos = None
for tos in terms_of_services:
    if tos.language == lang:
        current_tos = tos
```

plugins/quetz_tos/tests/test_quetz_tos.py
```diff
@@ -9,39 +9,81 @@ def plugins():
...
-    response = client.get('/api/tos')
+    params = {'language': 'EN'}
+    response = client.get('/api/tos', params=params)
+    print(response.json())
+    assert response.json()['filename'] == 'tos.txt'
+    assert response.json()['content'] == 'demo tos'
+
+def test_get_tos_fr(client, tos_file, tos):
+    params = {'language': 'FR'}
+    response = client.get('/api/tos', params=params)
```
martinRenou, line 85: Maybe you could try `response = client.get('/api/tos?lang=FR')` if this does not work.

plugins/quetz_tos/tests/test_quetz_tos.py
```diff
@@ -9,39 +9,81 @@ def plugins():
...
+    print(response.json())
     assert response.json()['filename'] == 'tos.txt'
```
martinRenou, line 87: The content of your FR terms of services file is `tos_fr.txt`, so you probably need to fix this line.

martinRenou (approved):
Thanks! To me this looks good.

@wolfv You'd probably want to have a look :)

martinRenou:
I am just a bit skeptical about the signing logic now.

Because it's one `TermsOfServices` object per-language, it's one `id` per-language. So when someone signs a specific version of the terms of services, they sign it for the language they read.
It's probably fine though?

janjagusch:
https://github.com/mamba-org/quetz/pull/552 got merged, so can we close this, @martinRenou, @MichaelKora?

PR #551, opened 2022-07-25
Remove package from repodata.json

### LINKED ISSUE #475: Delete package doesn't rebuild repodata
**Scenario:**
Published a package with abi break without bumping major version. So delete this version of the package and fix the version in the background.

**Problem:**
When the package is deleted from the server it can still be found in the repodata.json so the installation step gives you the error:
```
Downloading  [=======================================================================================] (00m:00s)    2.02 MB/s
Extracting   [>                                                                                                   ] (--:--)
ERROR   File not valid: file size doesn't match expectation "/home/user/micromamba/pkgs/ros-noetic-removed-package-4.1.0-h3fd9d12_0.tar.bz2"
        Expected: 854962
File size validation error.
Downloading  [=======================================================================================] (00m:00s)    2.02 MB/s
Extracting   [>                                                                                                   ] (--:--)
ERROR   Found incorrect download: ros-noetic-removed-package. Aborting
```

**Expected behavior:**
When the delete request is processed, the repodata.json should also be rebuilt.

brichet:
This PR cleans the `repodata.json` files in all platforms which was containing the removed package.

Fix #475

### DIFF
--- a/quetz/tests/api/test_main_packages.py
+++ b/quetz/tests/api/test_main_packages.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from typing import BinaryIO
 
@@ -10,6 +11,7 @@ from quetz.config import Config
 from quetz.db_models import ChannelMember, Package, PackageMember, PackageVersion
 from quetz.errors import ValidationError
 from quetz.pkgstores import PackageStore
+from quetz.tasks.indexing import update_indexes
 
 
 @pytest.mark.parametrize("package_role", [OWNER, MAINTAINER, MEMBER])
@@ -68,6 +70,14 @@ def test_delete_package_versions_with_package(
 
     assert package_version.package_name == public_package.name
 
+    update_indexes(dao, pkgstore, public_channel.name)
+
+    package_filenames = [
+        os.path.join(version.platform, version.filename)
+        for version in public_package.package_versions  # type: ignore
+    ]
+    init_files = sorted(pkgstore.list_files(public_channel.name))
+
     response = auth_client.delete(
         f"/api/channels/{public_channel.name}/packages/{public_package.name}"
     )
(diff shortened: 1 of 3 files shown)


wolfv:
nice, thanks for this PR!

I think we should add tests to make sure the two issues are properly fixed.

brichet:
Sure.
In fact I'm not sure there are already some tests on `repodata.json` content.

wolfv:
@brichet it's easiest if you install `pre-commit` and then do `pre-commit run --all` and it will fix everything for you with regards to linting.

brichet:
> @brichet it's easiest if you install `pre-commit` and then do `pre-commit run --all` and it will fix everything for you with regards to linting.

Thanks

brichet:
As there are no test on `repodata.json` file content yet, don't you think we can do that in another PR.
It would include some tests on creation, modification and deletion.

wolfv:
We can, but I would like to put it very high on the priority list.

quetz/tests/api/test_main_packages.py
```diff
@@ -85,9 +95,12 @@ def test_delete_package_versions_with_package(
 
     assert len(versions) == 0
 
-    files = pkgstore.list_files(public_channel.name)
+    for filename in package_filenames:
+        init_files.remove(filename)
```
wolfv, line 99: this should not happen when the subdir is `noarch` because otherwise the client thinks that the channel does not exist.

brichet: I'm not sure to understand.
This code removes the package filenames from the list of channel files (`repodata.json`, `index.html`...), to compare the files list before an after removing a package (only the package archive should have been removed).

brichet:
> We can, but I would like to put it very high on the priority list.


OK, I'm starting updating the tests anyway, it can be on that PR.

quetz/tests/api/test_main_packages.py
wolfv: Sorry, I thought that you were deleting empty repodata.json files. I should've looked more carefully.

wolfv: > Clean repodata.json from empty platform when deleting a package

That was the commit message that confused me :)

brichet: No problem, thanks for review

PR #552, opened 2022-07-27
Implement multiple languages support for the TermsOfServices

martinRenou:
Taking over #543 and making some changes in the original implementation.


wolfv:
@martinRenou for the linter, you can run `pre-commit run --all`

martinRenou:
Thanks!

plugins/quetz_tos/tests/test_quetz_tos.py
```diff
@@ -42,8 +43,14 @@ def test_tos_upload_by_owner(client, owner_user):
...
-    assert response.json()['filename'] == 'tos.txt'
-    assert response.json()['content'] == 'demo tos'
+
+    assert response.json()['files'][0]['language'] == 'EN'
+    assert response.json()['files'][0]['filename'] == 'tos_en.txt'
+    assert response.json()['files'][0]['content'] == 'demo tos en'
+
+    assert response.json()['files'][1]['language'] == 'FR'
+    assert response.json()['files'][1]['filename'] == 'tos_fr.txt'
+    assert response.json()['files'][1]['content'] == 'demo tos fr'
```
hbcarlos, line 53: Should we modify the GET to receive the language as a parameter? Otherwise, we are fetching all the languages on each request, and most clients are interested in just one language.

Something like `GET http://127.0.0.1:8000/api/tos?lang=EN`

For this REST end-point:
https://github.com/mamba-org/quetz/blob/1051658a1fe9fbbc1c02444aa3097307a1395d99/plugins/quetz_tos/quetz_tos/api.py#L38

hbcarlos (commented):
Thanks, @martinRenou! Looks good.
I'm thinking that maybe we should modify the `GET api/tos` end-point to receive the language as a parameter.

plugins/quetz_tos/tests/test_quetz_tos.py
martinRenou: Thanks!

I was thinking it was more convenient to fetch all languages at once, this way front-ends have the knowledge of which languages are available and can display a selection dropdown or something.

That being said we can probably reply with all the language versions if no language parameter is provided, and if a language is specified we can reply with only this language. I will push some changes for this :)

martinRenou: I pushed another commit adding a language parameter :)

hbcarlos (approved):
Thanks, @martinRenou!

PR #555, opened 2022-07-29
Fixes issues and tests on indexes

brichet:
This PR :
- updates indexes after each way to upload or delete a package
- adds some tests on update indexes function
- adds tests on duplicate upload, with and without `force` flag
- adds some tests on `repodata.json` file after deleting a package


wolfv:
Thank you @brichet this looks great!

Issue #539, opened 2022-05-31
Quetz frontend lacking Content-Type header for api-keys


riccardoporreca:
The issue can be easily reproduced using Docker image `mambaorg/quetz:v0.4.4`, as follows

```bash
echo '                       
[general]
frontend_dir = "/quetz-frontend"

[sqlalchemy]
database_url = "sqlite:///./quetz.sqlite"

[session]
# openssl rand -hex 32
secret = "b72376b88e6f249cb0921052ea8a092381ca17fd8bb0caf4d847e337b3d34cf8"
https_only = false
' > ~/quetz-test/config.toml
```

```bash
docker run --rm -v ~/quetz-test:/quetz-deployment mambaorg/quetz:v0.4.4 quetz create --dev /quetz-deployment
docker run --rm -v ~/quetz-test:/quetz-deployment -p 8000:8000 mambaorg/quetz:v0.4.4
```

- Browse to http://localhost:8000/api/dummylogin/alice
- Browse to http://localhost:8000/api-keys click "Request API Key" => no key is displayed
 
Inspecting the request hints at the issue
![image](https://user-images.githubusercontent.com/13663564/171300158-199d2b2e-a137-43c1-9f8c-903368c2a345.png)
![image](https://user-images.githubusercontent.com/13663564/171301068-a163253a-9649-41d3-a13b-50e98305ee0a.png)
reside in the lack of `Content-Type` header, which is indeed not specified (although commented out) in

https://github.com/mamba-org/quetz/blob/d63e2f155aa75dfe852b70674b01ab7732d234c2/quetz_frontend/src/views/ApiKeysView.vue#L49-L54

I tested locally by uncommenting the line, and the issue is indeed solved.

brichet:
Thanks @riccardoporreca for pointing out this issue and providing the means to fix it.
It should be fixed with https://github.com/mamba-org/quetz/pull/557

wolfv:
Hi @riccardoporreca thanks for pointing this out. Actually, we do not really maintain the frontend in this repository anymore. The new one is at https://github.com/mamba-org/quetz-frontend

riccardoporreca:
Thanks @wolfv for the pointer, I did see that repository but was not sure whether that was just version relatated somehow to JupyterLab, since it is a fork of https://github.com/jupyterlab-contrib/jupyterlab-app-cookiecutter :)

[...]

PR #557, opened 2022-08-16
Fix/several typo errors

brichet:
This PR fixes some little typo errors, including #539

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Run test upload using quetz-client. (For testing purposes, an API key is created
 
 ```bash
 export QUETZ_API_KEY=E_KaBFstCKI9hTdPM7DQq56GglRHf2HW7tQtq6si370
-quetz-client upload http://localhost:8000/api/channels/channel0 xtensor/linux-64/xtensor-0.16.1-0.tar.bz2 xtensor/osx-64/xtensor-0.16.1-0.tar.bz2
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -77,7 +77,7 @@ Run test upload using quetz-client. (For testing purposes, an API key is created
 
 ```bash
 export QUETZ_API_KEY=E_KaBFstCKI9hTdPM7DQq56GglRHf2HW7tQtq6si370
-quetz-client http://localhost:8000/api/channels/channel0 xtensor/linux-64/xtensor-0.16.1-0.tar.bz2 xtensor/osx-64/xtensor-0.16.1-0.tar.bz2
+quetz-client upload http://localhost:8000/api/channels/channel0 xtensor/linux-64/xtensor-0.16.1-0.tar.bz2 xtensor/osx-64/xtensor-0.16.1-0.tar.bz2
```
wolfv, line 80: thanks! 

I think we never released the version of the `quetz-client` where the `upload` subcommand is required. Do you think we can add a backwards-compatible "upload-without-subcommand" to the quetz-client?

brichet: I didn't find a way using `argparse`.
we can set a default `subpasers`, but it's working only if there is no following arguments (which is not the case when uploading a package).
The only way I found is to add it manually in sys.argv, like : 

``` python
if len(sys.argv) > 1 and sys.argv[1] not in ALLOWED_SUBCOMMANDS:
    sys.argv.insert(1, "upload")
```

brichet: Finally I implemented https://stackoverflow.com/a/46964652/15860586, the same idea as above but without changing `sys.argv`, only using `argparse`, which is better.

PR #558, opened 2022-08-18
no file or directory mandatory

### LINKED ISSUE #459: Deployment directory shouldn't be mandatory
It's easily possible to run a quetz instance by setting the environment variables QUETZ_SQLALCHEMY_DATABASE_URL, QUETZ_SESSION_SECRET and it's possible to configure an external backend in this way too. Which is extremely useful for running quetz inside a Docker container. 

But even if the config is set using environment variables and pointing to external resources quetz still insists on:
- a deployment directory
- an (empty) config.toml
- an (empty) channels directory

This is pretty confusing, because in order to run quetz  in e.g. Docker or via docker-compose you need to create those empty files and directories that serve no purpose at all and make it quite non-obvious how quetz is actually configured or where it's storing packages.

brichet:
This PR should fix #459

The config file is now not mandatory, but the server will not start if one of the variables `$QUETZ_SQLALCHEMY_DATABASE_URL` and `$QUETZ_SESSION_SECRET` is missing.
The deployment directory is not mandatory either if the store is not a `LocalStore`. In fact if any other available store has a variable defined.

This also adds some basic tests to ensure server does not start without minimal settings.

**MISSING TESTS** : make sure the server starts correctly without settings file but with environment variables. Haven't found a way to kill it cleanly, so all following tests will fail with `Address already in use`.

### DIFF
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -298,7 +298,6 @@ def _is_deployment(base_dir: Path):
     config_file = base_dir.joinpath("config.toml")
     if (
         base_dir.exists()
-        and config_file.exists()
         and base_dir.joinpath("channels").exists()
     ):
         config = Config(str(config_file.resolve()))
@@ -474,9 +473,6 @@ def create(
 def _get_config(path: Union[Path, str]) -> Config:
     """get config path"""
     config_file = Path(path) / 'config.toml'
-    if not config_file.exists():
-        typer.echo(f'Could not find config at {config_file}')
-        raise typer.Abort()
 
     config = Config(str(config_file.resolve()))
     if not os.environ.get(_env_prefix + _env_config_file):
--- a/quetz/cli.py
+++ b/quetz/cli.py
@@ -297,10 +297,7 @@ def _fill_test_database(db: Session) -> NoReturn:
 
 def _is_deployment(base_dir: Path):
     config_file = base_dir.joinpath("config.toml")
-    if (
-        base_dir.exists()
-        and base_dir.joinpath("channels").exists()
-    ):
+    if base_dir.exists() and base_dir.joinpath("channels").exists():
         config = Config(str(config_file.resolve()))
         with working_directory(base_dir):
             if not database_exists(config.sqlalchemy_database_url):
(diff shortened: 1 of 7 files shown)


brichet:
https://github.com/mamba-org/quetz/pull/558/commits/ad6c7e6aeed45edeb592a1a73b069dc3736f0be0 allows to create a local storage without configuration file.

**WARNING** : If the database is sqlite, the database url includes the local path to the database file. This database URL can be set from the configuration file or the environmental variable, and currently we don't know where it comes from.
In that case, if a configuration file exists, the database path will be relative to that config file path. If the configuration file does not exist, the database path will be relative to the current working directory.

This means that if the database url is set in environmental variable **and** is relative to CWD **and** a configuration file exists in an other directory **=>** it will fail to start.

quetz/cli.py
```diff
@@ -463,7 +482,8 @@ def create(
     config = Config(str(config_file))
 
     deployment_folder.joinpath('channels').mkdir(exist_ok=True)
-    with working_directory(path):
+
+    with working_directory(db_path):
```
hbcarlos, line 486: Could we use the `deployment_folder` as a working directory if it exists?
I'm testing this PR with a local store, so I need to specify a working directory, and instead of using this directory to store the SQLite database, it creates the database relative to CWD. I think it would be better to reuse the `deployment_folder` since we have it, and if there is not a `deployment_folder` then we use CWD.

quetz/testing/fixtures.py
```diff
@@ -212,18 +225,20 @@ def config(config_str, config_dir, test_data_dir):
 
     Config._instances = {}
     config = Config()
+    print(config_dir)
```
hbcarlos, line 228: Leftover?

hbcarlos (approved):
Thanks, @brichet! I left two minor comments.
Everything else looks good to me.

quetz/cli.py
brichet: I don't think so, because the `$QUETZ_SQLALCHEMY_DATABASE_URL` includes the relative path to the database. It would be broken if it startd from the deployment_folder.

The best way to keep it generic is to let the user define it in the database URL:

```
export QUETZ_SQLALCHEMY_DATABASE_URL="sqlite:///./deployment_folder/quetz.sqlite"
```

hbcarlos: Ooh, that's right.

PR #568, opened 2022-10-26
Fix update route in SQL Authenticator

simonbohnen:


plugins/quetz_sql_authenticator/quetz_sql_authenticator/api.py
```diff
@@ -95,7 +104,7 @@ def _update(
             status_code=status.HTTP_404_NOT_FOUND,
             detail=f"User {username} not found",
         )
-    credentials.password = _calculate_hash(password)
+    credentials.password_hash = _calculate_hash(password)
```
simonbohnen, line 107: This is the critical change.

plugins/quetz_sql_authenticator/quetz_sql_authenticator/api.py
```diff
@@ -72,8 +73,16 @@ def _create(
...
+    try:
+        db.commit()
+    except IntegrityError:
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=f"""
+            A database error occured.
+            This is most likely due to the user {username} already existing.
+            """,
+        )
```
janjagusch, line 85: can we make this error message a bit more detailed?

i suggest the following:

1. Parse the integrity error
2. Figure out if the integrity error is due to a name conflict
3. If yes, return a corresponding error message (something like 'username already exists.')
4. If no, return the integrity error message (something like: 'a database error occurred: {integrity error message}')

janjagusch (changes requested):
nice! just one small change request, then we can merge this. 👍

quetz/testing/fixtures.py
```diff
@@ -147,8 +148,35 @@ def expires_on_commit():
...
+    rollback_support_marker = request.node.get_closest_marker(
+        "support_sqlalchemy_rollback"
+    )
+
+    if (
+        rollback_support_marker is not None
+        and rollback_support_marker.args[0] is True
+        and auto_rollback
+    ):
+        # See https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites # noqa
```
janjagusch, line 163: Do you need the `# noqa` at the end of the line?

quetz/testing/fixtures.py
```diff
@@ -147,8 +148,35 @@ def expires_on_commit():
...
+
+    rollback_support_marker = request.node.get_closest_marker(
+        "support_sqlalchemy_rollback"
+    )
+
+    if (
+        rollback_support_marker is not None
+        and rollback_support_marker.args[0] is True
+        and auto_rollback
+    ):
```
janjagusch, line 162: Maybe add a comment about the motivation behind this conditional.

quetz/testing/fixtures.py
```diff
@@ -147,8 +148,35 @@ def expires_on_commit():
 
 
 @pytest.fixture
-def db(session_maker, expires_on_commit):
+def db(session_maker, expires_on_commit, auto_rollback, request):
```
janjagusch, line 151: Thinking out loud: Instead of changing this fixture in `quetz/testing/fixtures`, would it make sense to redefine it in the `conftest.py` of our plugin?

janjagusch (commented):
nice work!

quetz/testing/fixtures.py
simonbohnen: I don't think it matters. I can document this a bit more, and then it might even be useful for other plugins.

janjagusch: Motivation: What if we wanted to move the plugin into a separate repository? In that case, we can't just edit fixtures defined inside `quetz`. Our only possibility would be to redefine the fixture, right?

It would be nice to have the plugin in a state where we can quickly move it into a separate repository, which I why I suggested this.

simonbohnen: That makes sense 👌🏼

janjagusch (approved):
awesome job!

PR #569, opened 2022-10-28
Fix sqlauth error parsing

simonbohnen:


janjagusch (commented):
thanks for your work!

plugins/quetz_sql_authenticator/quetz_sql_authenticator/api.py
```diff
@@ -77,7 +77,8 @@ def _create(
         db.commit()
     except IntegrityError as e:
         db.rollback()
-        if "UNIQUE constraint failed: credentials.username" in str(e):
+        if "unique" in str(e).lower() and "INSERT INTO credentials" in str(e):
```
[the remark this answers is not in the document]
simonbohnen: I agree, not leaking the error makes sense 👍🏼

plugins/quetz_sql_authenticator/quetz_sql_authenticator/api.py
```diff
@@ -136,5 +132,23 @@ def _delete(
...
+            f"""
+        quetz-sql-authenticator encountered the following error \
+        while trying to commit changes to the database:
+        {e}
+        """
+        )
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="An internal error occured.",
+        )
```
janjagusch, line 154: Ok, with what we've discussed before, do we still need to safe guard the `db.commit()` call? I think not because:

1. When we raise an error inside a route, we send an internal server error response back to the client (same as in this function).
2. By raising the error, we get the entire traceback, not only the error message (better than this function).

So, it seems like there is no harm in raising a potential error here, which would make this function redundant.

Sorry for going back and forth on this topic. I didn't consider the implications when I first proposed this.

simonbohnen: Agreed. It seems like FastAPI or whatever part handles the error just returns "Internal Server Error" as HTTP and logs the error 👍🏼

plugins/quetz_sql_authenticator/quetz_sql_authenticator/api.py
```diff
@@ -10,6 +12,7 @@
 from .db_models import Credentials
 
 router = APIRouter()
+logger = logging.getLogger('quetz')
```
janjagusch, line 15: We can remove the logger again.

janjagusch (commented):
almost there!

PR #570, opened 2022-10-28
Add 'dependencies' to PR labels

janjagusch:
This should make CI pass in `dependabot` PRs, like https://github.com/mamba-org/quetz/pull/566.

### DIFF
--- a/.github/workflows/enforce-label.yml
+++ b/.github/workflows/enforce-label.yml
@@ -2,7 +2,7 @@ name: Enforce PR label
 
 on:
   pull_request:
-    types: [labeled, unlabeled, opened, edited, synchronize]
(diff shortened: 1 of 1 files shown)


janjagusch:
I just realized I changed this in the wrong location and that the labels are hard-coded into the action: https://github.com/jupyterlab/maintainer-tools/blob/9309d307fc04b19ed63e5ec8e353d93aa836b727/.github/actions/enforce-label/action.yml#L14

wolfv:
Yeah I think I went down this rabbit hole once as weel. Anyways, we're trying to get rid of the frontend here and focus on `mamba-org/quetz-frontend` instead #565

PR #574, opened 2022-11-16
Remove `httpx` pin, fix tests, & test plugins individually

simonbohnen:
This PR removes the pin on `httpx`, which was causing issues with newer versions of `starlette` which require `httpx>=0.22.0`.
See https://github.com/conda-forge/quetz-feedstock/issues/24.

- Fixed the tests which required changes to work with `httpx>=0.22.0`
- Changed plugin testing to not abort when tests have failed for one plugin

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
           echo hello
 
       - name: Step 2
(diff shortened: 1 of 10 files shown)


janjagusch (approved):
Thank you!

PR #586, opened 2023-01-31
use use raise_for_status instead of printing empty response.json()

DerThorsten:
when successfully creating a channel, the quetz-client  prints None
```
quetz-client create-channel emscripten-forge-bootstrap 
None
```
because we do `print(response.json())`.
We should rather do `response.raise_for_status()` which will throw an error when we get a error indicating HTML return code.


janjagusch:
Thanks for your PR, @DerThorsten. We are working on a new Quetz client (https://github.com/mamba-org/quetz/pull/589), which will make this code obsolete. Would you mind if I close the PR in favor of the new client?

janjagusch:
We have removed the old quetz client, so I'm closing this PR.

PR #594, opened 2023-02-09
Fix upload endpoint

simonbohnen:
Modify the `upload` endpoint to include
- updating the `channeldata`
- reindexing the channels

Also replace the incorrect `NoReturn` type hints with `None`. Functions which return without a value should be typed as `None`, `NoReturn` is meant for functions which [never finish](https://docs.python.org/3/library/typing.html#typing.NoReturn).


wolfv:
perfect, I think now we can also update `mypy`! :)

PR #602, opened 2023-02-16
Remove xattr as hard dependency

SimonBohnenQC:
`xattr` is not a hard dependency and should thus be removed.

Having `xattr` as a dependency leads to the following problem:
- xattr is not available on windows via conda-forge, see [here](https://anaconda.org/conda-forge/xattr)
- `quetz` is a noarch package, see [here](https://github.com/conda-forge/quetz-feedstock/blob/83c8cffe821002435219cc1e9354cfdeccf4a2d3/recipe/meta.yaml#L13)
- the feedstock only builds using linux, so the problem isn't noticed there.
- but when building a new recipe, e.g. the one for the [`quetz-sql-authenticator`](https://github.com/conda-forge/staged-recipes/pull/22038), the build is also tested on Windows and fails, see [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=659440&view=logs&j=240f1fee-52bc-5498-a14a-8361bde76ba0&t=2fb5b0a7-737c-5f5c-8f3f-f6db6174bacf&l=326)


.github/workflows/ci.yml
```diff
@@ -162,3 +162,4 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
         env:
           RH_BRANCH: ${{ steps.branch-name.outputs.current_branch }}
+          RH_REPO: ${{ github.repository }}
```
janjagusch, line 165:
```suggestion
          RH_REPOSITORY: ${{ github.repository }}
```

simonbohnen: Did this manually 👌🏼

janjagusch (approved):

PR #609, opened 2023-02-18
Fix unbound variables

### LINKED ISSUE #606: First level attribute can be unbound
There is a change that you never enter the while loop, which makes the `first_level` attribute undefined.

https://github.com/mamba-org/quetz/blob/efcfdb8259a44b6765f2962a29047889bc70f48e/quetz/config.py#L395-L408

This can cause the following error.

```
UnboundLocalError: cannot access local variable 'first_level' where it is not associated with a value
```

I feel like this should have been caught by some pre-commit hook (e.g. `mypy`).

janjagusch:
Depends on #607 
Closes #606 

This is a first attempt to fix all possibly unbound variables in the code. I've decided to use the most straightforward fix I could think of. But I feel like some of the code has structural problems and that a refactoring might be in order.


plugins/quetz_repodata_zchunk/tests/test_main.py
```diff
@@ -142,7 +142,7 @@ def test_repodata_zchunk(
             with open('repodata.json') as f:
                 repodata_unzck = f.read()
 
-            assert repodata == repodata_unzck  # NOQA
+            assert repodata == repodata_unzck  # NOQA # type: ignore
```
janjagusch, line 145: This seems utterly broken. The `repodata` variable isn't defined anywhere. Do we run the plugin tests in CI?

quetz/cli.py
```diff
@@ -546,6 +546,7 @@ def start(
             )
             raise typer.Abort()
 
+    supervisor_process = None
```
janjagusch, line 549: I feel like there should be a more elegant pattern to do the same thing.

quetz/dao.py
```diff
@@ -860,7 +861,9 @@ def create_version(
                     )
                     if is_newer:
                         break
-                version_order = v.version_order if is_newer else v.version_order + 1
+                version_order = (
+                    v.version_order if is_newer else v.version_order + 1  # type: ignore
+                )
```
janjagusch, line 866: This needs to be improved.

quetz/utils.py
```diff
@@ -160,6 +162,8 @@ def apply_custom_query(search_type, db, keywords, filters):
                 each_keyword_condition = collate(Channel.name, "und-x-icu").notlike(
                     f'%{negation_argument}%'
                 )
+            else:
+                raise KeyError(search_type)
```
janjagusch, line 166: We should probably use dictionaries here.

plugins/quetz_repodata_zchunk/tests/test_main.py
simonbohnen: We do, i can check this tomorrow 👍🏼

simonbohnen: Probably the `if` branch is never run, and the `else` branch doesn't contain any asserts 🤷🏼‍♂️

PR #621, opened 2023-04-10
Fix pyright errors

janjagusch:


quetz/dao.py
```diff
@@ -651,6 +653,8 @@ def get_channel_member(self, channel_name, username):
 
     def create_channel_member(self, channel_name, new_member):
         user = self.get_user_by_username(new_member.username)
+        if user is None:
```
janjagusch, line 656: This error should probably be caught and returned as a FastAPI HTTP error.

quetz/dao.py
```diff
@@ -681,6 +685,8 @@ def get_package_member(self, channel_name, package_name, username):
 
     def create_package_member(self, channel_name, package_name, new_member):
         user = self.get_user_by_username(new_member.username)
+        if user is None:
```
janjagusch, line 688: This error should probably be caught and returned as a FastAPI HTTP error.

quetz/main.py
```diff
@@ -937,7 +947,12 @@ def post_channel_member(
         channel_member.role = new_member.role
         db.commit()
     else:
-        dao.create_channel_member(channel.name, new_member)
+        try:
+            dao.create_channel_member(channel.name, new_member)
+        except KeyError as error:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail=error.args[0]
+            ) from error
```
janjagusch, line 955: Thinking about it: This changes user-facing behavior. Maybe let's put this in a separate PR.

PR #622, opened 2023-04-10
Add endpoints for health checks

### LINKED ISSUE #620: Add health endpoint

janjagusch:
Closes #620.

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -253,7 +253,7 @@ def dummy_login(
     if user is None:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
-            detail=f"User '{username}' not found.",
+            detail=f"User '{username}' not found",
         )
 
     logout(session)
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -250,6 +250,8 @@ def dummy_login(
     username: str, dao: Dao = Depends(get_dao), session=Depends(get_session)
 ):
     user = dao.get_user_by_username(username)
+    if user is None:
+        raise HTTPException(status_code=404, detail=f"User '{username}' not found.")
 
     logout(session)
     session["user_id"] = str(uuid.UUID(bytes=user.id))
@@ -438,6 +440,11 @@ def delete_user(
     auth: authorization.Rules = Depends(get_rules),
 ):
     user = dao.get_user_by_username(username)
+
+    if not user or not user.profile:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail=f"User {username} not found"
+        )
 
     auth.assert_delete_user(user.id)
     dao.delete_user(user.id)
@@ -1105,6 +1112,11 @@ def delete_package_version(
     version = dao.get_package_version_by_filename(
         channel_name, package_name, filename, platform
     )
+
+    if not version or not version.package:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Package version not found"
+        )
 
     auth.assert_package_delete(version.package)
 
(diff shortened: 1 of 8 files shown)


quetz/main.py
```diff
@@ -1839,4 +1839,14 @@ def serve_channel_index(
     return serve_path("index.html", channel, accept_encoding, session, dao)
 
 
+@app.get("/health/ready", status_code=status.HTTP_200_OK)
+def is_ready(task: Task = Depends(get_tasks_worker)):
+    return {}
```
janjagusch, line 1844: I'm unsure about the correct definition and readiness and liveness for Quetz. [But it seems like when a task worker can be instantiated](https://github.com/mamba-org/quetz/blob/87ec595410ba39de246bdda221f58735cb3981a0/quetz/deps.py#L84-L91), the whole service should be available.

quetz/tests/test_health.py
```diff
@@ -0,0 +1,11 @@
+from fastapi.testclient import TestClient
```
borchero, line 1: Do you know if we can easily add a test case where the health endpoints don't return 200?

borchero (commented):
Thanks for implementing @janjagusch! I'd rename the endpoints to be consistent with the [Kubernetes API health endpoints](https://kubernetes.io/docs/reference/using-api/health-checks/), i.e.

- `/health/ready` -> `/readyz`
- `/health/live` -> `/livez`

quetz/tests/test_health.py
janjagusch: I think with the FastAPI test client, this will be difficult. We would probably need to set up a proper integration test with a running Quetz server instance.

ivergara: It is possible to [override dependencies for testing](https://fastapi.tiangolo.com/advanced/testing-dependencies/), that's kind of the whole point of Dependency Injection. Now the question is exactly how to write the function. Then, that'd illuminate to properly implement the endpoint and the tests.

quetz/main.py
```diff
@@ -1839,4 +1839,14 @@ def serve_channel_index(
     return serve_path("index.html", channel, accept_encoding, session, dao)
 
 
+@app.get("/health/ready", status_code=status.HTTP_200_OK)
```
ivergara, line 1842:
```suggestion
@app.get("/health/live", status_code=status.HTTP_200_OK)
@app.get("/health/ready", status_code=status.HTTP_200_OK)
```

Since both endpoints do the same I'd do this instead.

janjagusch:
> Thanks for implementing @janjagusch! I'd rename the endpoints to be consistent with the [Kubernetes API health endpoints](https://kubernetes.io/docs/reference/using-api/health-checks/), i.e.
> 
> * `/health/ready` -> `/readyz`
> * `/health/live` -> `/livez`

@borchero, have you ever seen this convention anywhere outside the k8s API? Most articles about health check endpoints suggest a name scheme similar to the one I implemented here (e.g. [here](https://testfully.io/blog/api-health-check-monitoring/#api-health-check-endpoint-types)). I must say I prefer the current naming scheme. :)

quetz/tests/test_health.py
janjagusch: Setting up an integration test will be complicated, so I wouldn't include that in this PR.

borchero:
@janjagusch I've seen it a bunch of times (see https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f which would also imply it's used in quite a few places) but I don't have a strong opinion. I've also seen the `/health/xxx` naming :smile:

borchero (approved):

ivergara:
Both can be created and the discussion is over!

PR #623, opened 2023-04-14
Add sampler profiling

ivergara:
This PR leverages `pyinstrument` to provide sampling profiling to all endpoints exposed by `quetz`.

For this to happen, the `enable_sampling` configuration of the section `profiling` must be set to `True` and then adding`profiling=true` (or any value that evaluates to `True`) as a query parameter. Once this is done, the response will not be the expected one, but an HTML view of the profiling done by `pyinstrument`.

### DIFF
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -225,6 +225,16 @@ You can configure the limits (quota) on the size of uploaded packages for each c
 
 :channel_quota: maximum total size (in bytes) of packages uploaded to the channel
 
+``profiling`` section
+^^^^^^^^^^^^^^^^^^^^^
+
+Quetz provides instrumentation for profiling its endpoints.
+
+:enable_sampling: enables sampling profiling by providing the query parameter `profile=true` (or any truth value) to the request.
+When active and provided the query parameter, the returned response will be highjacked to provide an HTML version of the profile output.
+
+:interval_seconds: sampling interval in seconds. If not set, it has a default value of `0.001`.
+
 Environment
 -----------
 
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -230,8 +230,7 @@ You can configure the limits (quota) on the size of uploaded packages for each c
 
 Quetz provides instrumentation for profiling its endpoints.
 
-:enable_sampling: enables sampling profiling by providing the query parameter `profile=true` (or any truth value) to the request.
-When active and provided the query parameter, the returned response will be highjacked to provide an HTML version of the profile output.
+:enable_sampling: enables sampling profiling by providing the query parameter `profile=true` (or any truth value) to the request. When active and provided the query parameter, the returned response will be highjacked to provide an HTML version of the profile output.
 
 :interval_seconds: sampling interval in seconds. If not set, it has a default value of `0.001`.
 
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -165,28 +165,24 @@ class CondaTokenMiddleware(BaseHTTPMiddleware):
 app.add_middleware(CondaTokenMiddleware)
 
 
-if config.profiling_enable_sampling:
-    try:
-        from pyinstrument.profiler import Profiler
-    except ModuleNotFoundError:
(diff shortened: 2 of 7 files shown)


AndreasAlbertQC:
@ivergara could you add some documentation on the new config setting [here](https://github.com/mamba-org/quetz/blob/main/docs/source/deploying/configuration.rst)?

quetz/main.py
```diff
@@ -163,6 +164,31 @@ async def dispatch(self, request, call_next):
 
 app.add_middleware(CondaTokenMiddleware)
 
+
+if config.profiling_enable_sampling:
+    try:
+        from pyinstrument.profiler import Profiler
+    except ModuleNotFoundError:
+        logger.warning("'pyinstrument' was not found, proceeding without profiling.")
+    else:
```
janjagusch, line 173: Should it be possible to enable sampling without having `pyinstrument` installed? Shouldn't we just raise hard when that's the case?

janjagusch (commented):
Thanks for this PR!

Something about your changes seems to break CI ([it was running successfully on the default branch this morning](https://github.com/mamba-org/quetz/actions/runs/4665660541/jobs/8443137314)). Could you please take a look?

quetz/main.py
ivergara: Yes, is an alternative. Since you suggest it and there is a 👍 I'll remove the try/except.

ivergara:
@janjagusch @AndreasAlbertQC a few options
1. should I add the option to activate profiling in the dev configuration?
2. How do you recommend testing this? Or is enough manual QA since this is basically for dev purposes?

janjagusch:
> 1. should I add the option to activate profiling in the dev configuration?

You mean in [this file](https://github.com/mamba-org/quetz/blob/main/dev_config.toml)? I don't have a strong preference. What about adding the section but setting it to `false`?

> 2. How do you recommend testing this? Or is enough manual QA since this is basically for dev purposes?

I'm unsure how easily we could integrate this in the current testing suite. If it's too much work, then I suggest we leave it untested for now. Let's do a manual demonstration together, so I can see that it's working.

docs/source/deploying/configuration.rst
```diff
@@ -225,6 +225,16 @@ You can configure the limits (quota) on the size of uploaded packages for each c
 
 :channel_quota: maximum total size (in bytes) of packages uploaded to the channel
 
+``profiling`` section
+^^^^^^^^^^^^^^^^^^^^^
+
+Quetz provides instrumentation for profiling its endpoints.
+
+:enable_sampling: enables sampling profiling by providing the query parameter `profile=true` (or any truth value) to the request.
+When active and provided the query parameter, the returned response will be highjacked to provide an HTML version of the profile output.
```
janjagusch, line 234: This should fix the docs CI step.

```suggestion
:enable_sampling: enables sampling profiling by providing the query parameter `profile=true` (or any truth value) to the request. When active and provided the query parameter, the returned response will be highjacked to provide an HTML version of the profile output.
```

ivergara:
> > 1. should I add the option to activate profiling in the dev configuration?
> 
> You mean in [this file](https://github.com/mamba-org/quetz/blob/main/dev_config.toml)? I don't have a strong preference. What about adding the section but setting it to `false`?

Indeed that file. Sure, I can change it to false.

 
> > 2. How do you recommend testing this? Or is enough manual QA since this is basically for dev purposes?
> 
> I'm unsure how easily we could integrate this in the current testing suite. If it's too much work, then I suggest we leave it untested for now. Let's do a manual demonstration together, so I can see that it's working.

That's why I suggested just going for manual QA only. I could see how to write a test that if profiling is enabled and the query parameter is given, the response is an HTML page instead of a JSON answer (probably via the content-type of the response). But this sounds like a very weak test.

AndreasAlbertQC:
> But this sounds like a very weak test.

True, but still better than no test, though. At least it will break if something in the setup changes (e.g. some unexpected breaking change in pyinstrument).

janjagusch:
> > But this sounds like a very weak test.
> 
> True, but still better than no test, though. At least it will break if something in the setup changes (e.g. some unexpected breaking change in pyinstrument).

I'm also in favor of this smoke test. After that, we should be good to merge.

PR #626, opened 2023-04-21
Add async friendly upload

ivergara:
Still a work in progress, but should be working in the local package store.

the `add_package` was not directly tested, only the `add_files`, but the upload doesn't use the `add_files`. Thus I added a very minimal test for `add_package` (with a bonus refactoring) and then went to turn it into an asynchronous function. `aiofiles` was already part of the dependencies but was never used until now, which provides a facade of synchronicity for FS local operations.

We can discuss some decisions, but first I wanted to make it work. Instead of saving the uploaded file with a random name and then renaming it, I'm writing it directly with the correct name. I tried the other way but didn't manage. Could give it another try if it's really critical.

Anyhow, would appreciate early feedback.

Edit:

I did end up keeping the original `add_package` because it's used in other areas and don't want to pollute the codebase with async/await keywords. Thus, added a separate async version for `add_package`. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204438786781707

### DIFF
--- a/quetz/pkgstores.py
+++ b/quetz/pkgstores.py
@@ -19,6 +19,7 @@ from threading import Lock
 from typing import IO, List, Tuple, Union
 
 import aiofiles
+import aioshutil
 import fsspec
 from tenacity import retry
 from tenacity.retry import retry_if_exception_type
@@ -350,7 +351,11 @@ class S3Store(PackageStore):
     async def add_package_async(
         self, package: File, channel: str, destination: str
     ) -> None:
-        raise NotImplementedError
+        with self._get_fs() as fs:
+            bucket = self._bucket_map(channel)
+            with fs.open(path.join(bucket, destination), "wb", acl="private") as pkg:
+                # use a chunk size of 10 Megabytes
+                aioshutil.copyfileobj(package, pkg, 10 * 1024 * 1024)
 
     def add_file(
         self, data: Union[str, bytes], channel: str, destination: StrPath
@@ -494,7 +499,11 @@ class AzureBlobStore(PackageStore):
     async def add_package_async(
         self, package: File, channel: str, destination: str
     ) -> None:
-        raise NotImplementedError
+        with self._get_fs() as fs:
+            container = self._container_map(channel)
+            with fs.open(path.join(container, destination), "wb") as pkg:
+                # use a chunk size of 10 Megabytes
+                aioshutil.copyfileobj(package, pkg, 10 * 1024 * 1024)
 
     def add_file(
         self, data: Union[str, bytes], channel: str, destination: StrPath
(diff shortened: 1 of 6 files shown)


quetz/pkgstores.py
```diff
@@ -614,13 +638,18 @@ def remove_channel(self, name):
         with self._get_fs() as fs:
             fs.rm(channel_path, recursive=True)
 
-    def add_package(self, package: File, channel: str, destination: str) -> None:
+    async def add_package(self, package: File, channel: str, destination: str) -> None:
```
janjagusch, line 641:
```suggestion
   def add_package(self, package: File, channel: str, destination: str) -> None:
```

quetz/pkgstores.py
```diff
@@ -465,13 +484,18 @@ def remove_channel(self, name):
...
```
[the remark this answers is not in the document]
ivergara: Yup, didn't commit that. Good catch!

ivergara:
> This is a great first attempt, and I like the general structure of the change.

Thanks! I've tried to be as noninvasive as possible. However, besides passing the tests, we don't know yet if it's solving the originating issue. 

>  I would be very interested in seeing implementations and tests for the cloud blob stores, especially GCS.

I'm also curious about how it'd look like.

Should we add the changes to activate testing against a GCS blob store in a separate PR?

janjagusch:
@ivergara, can you take care of the lint? then this should be good to go. 👍

ivergara:
> quetz/tests/test_pkg_stores.py::test_remove_dirs[local_store] PASSED     [ 23%]
> quetz/tests/test_pkg_stores.py::test_remove_dirs[gcs_store] PASSED       [ 23%]
> quetz/tests/test_pkg_stores.py::test_store_add_list_files[gcs_store] PASSED [ 23%]
> quetz/tests/test_pkg_stores.py::test_move_file[gcs_store] PASSED         [ 24%]
> quetz/tests/test_pkg_stores.py::test_create_channel_multiple_times[gcs_store] PASSED [ 25%]

The tests against GCS did pass 🚀! Thus, at least, behavior wise it's working. Will fix the linting issue and then we're ready to merge and do a manual stress test.

quetz/pkgstores.py
```diff
@@ -571,6 +604,7 @@ def __init__(self, config):
             token=self.token if self.token else None,
             cache_timeout=self.cache_timeout,
             default_location=self.region,
+            asynchronous=True,
```
janjagusch, line 607: Do we also have to pass the `asynchronous=True` falg to the other `fsspec`-based package stores?

ivergara: That's a good question and observation. My hunch is that it should, but I'll check and most likely add that.

PR #628, opened 2023-05-01
Avoid exposing the Postgres credentials

sbivol:
Quetz prints the database connection string on startup. This is unnecessary and potentially insecure, since the person reading Quetz logs is not necessarily authorized to manage Quetz.
This patch replaces the connection string with just the name of the database engine: SQLite, PostgreSQL, or whatever happens to be preceding the first slash (useful in case the user made a typo, or other DB engines become supported in the future).


wolfv (approved):

wolfv:
Thanks!

PR #635, opened 2023-05-31
Make upload routes consistent with each other

AndreasAlbertQC:
We currently have two very similar file upload routes:
1. `/channels/{channel_name}/packages/{package_name}/files/` (`post_file_to_package`)
2. `/channels/{channel_name}/upload/{file_name}` (`post_upload`)

The first one is currently well-tested, but the latter is not. This PR updates the testing so that more similar tests are applied to both routes. While doing this, I found that the second route does not check if uploaded file and package names are consistent, which the former route does. I also made this consistent now.

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -1371,7 +1371,6 @@ async def post_upload(
     )
 
     upload_hash = hashlib.sha256()
-
     body = TemporaryFile()
     async for chunk in request.stream():
         body.write(chunk)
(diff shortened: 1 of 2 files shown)


janjagusch:
Thanks for the clean-up, @AndreasAlbertQC! What do you think is the motivation behind having two upload routes? Would it make sense to deprecate one of them?

AndreasAlbertQC:
Great question @janjagusch, this was also my first instinct. I have not fully understood the differences between the routes, yet, so I cannot make that call reliably at the moment. 

Side note: There's actually also `/channels/{channel_name}/files/` (`post_file_to_channel`), but the implementation of that function already seems to be the same as `post_file_to_package`, so there the call might be easier.

wolfv (approved):

PR #639, opened 2023-06-01
multiple mirror sync with extensive include/exclude pattern matching

YYYasin19:
Hi,

this is a draft PR to check out how much effort it would be to
- allow quetz to mirror packages from multiple channels
- define more elaborate include/exclude list patterns on mirror channels

We've had the need for these features for a long time and, until now, worked around them with internal tooling.

## Reasoning
Mirror Servers are not only used for load distribution but also **access management** to packages.
There are multiple common cases where it might be important to have more control over the sync feature:
- Company provides access to a mirrored server that has package _versions_ that are permitted (e.g. because of outstanding security checks or licensing issues)
- A software company distributes IP (aka code) as packages to customers and needs to share this IP

## Changes
The needed changes to allow all these are:
- A more fine-grained filter on packages (e.g. `my_cool_package=1.5.*`)
- Sync should also _remove_ packages when they're not in the `includelist` anymore
- Allow a channel to mirror multiple channels

---
I'm happy to discuss other (maybe even easier) ways of implementing this! 😊


quetz/utils.py
```diff
@@ -39,6 +39,47 @@ def check_package_membership(package_name, includelist, excludelist):
     return True
 
 
+def _include_pattern_match(name, version, build, pattern):
```
wolfv, line 42: This looks a lot like a MatchSpec - I think we should just use that :)

That means we should either depend on conda or vendor the MatchSpec part from conda into quetz. I think there were also discussions on making MatchSpec standalone but not sure if they ever got anywhere.

YYYasin19: Nice catch!
We actually have `conda` as a dependency already because of `conda-build` and `mamba` apparently.

I'm now using MatchSpec and since it also works for _just_ the package name, we might be able to get rid of the additional fields `include_pattern_list` and just use the `include_list` again, making the transition even simpler.

YYYasin19:
rebased ✅

quetz/main.py
```diff
@@ -1798,14 +1801,24 @@ def serve_path(
...
-        repository = RemoteRepository(channel.mirror_channel_url, session)
+            # proxying packages only works for the first url in the list
+            redirect_url = channel.mirror_channel_urls[0]
+            return RedirectResponse(f"{redirect_url}/{path}")
+
+    # note: you can only proxy, when you have exactly one remote channel in your config
+    if channel.mirror_channel_urls and channel.mirror_mode == "proxy":
+        proxy_url = channel.mirror_channel_urls[0]
+        if len(channel.mirror_channel_urls) > 1:
+            logger.warning(
```
AndreasAlbertQC, line 1817: I think generally warnings about user-input in a webserver are a bit ineffective because the person who is doing something wrong never gets to see the warning. Should this just be a failure?

quetz/pkgstores.py
```diff
@@ -67,6 +67,10 @@ def support_redirect(self) -> bool:
 
     @abc.abstractmethod
     def create_channel(self, name):
+        """
+        create channel with given name, depending on backend,
```
AndreasAlbertQC, line 71:
```suggestion
        create backend resources for a channel with given name,
        e.g. create a directory or bucket
```

quetz/rest_models.py
```diff
@@ -7,11 +7,12 @@
 import uuid
 from datetime import date, datetime
 from enum import Enum
-from typing import Generic, List, Optional, TypeVar
+from typing import Annotated, Dict, Generic, List, Optional, TypeVar, Union
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 T = TypeVar('T')
+URLType = Annotated[str, Field(pattern="^(http|https)://.+")]
```
AndreasAlbertQC, line 15: quetz currently supports python versions as far back as `3.7`. `Annotated` is only available in 3.9+

quetz/rest_models.py
```diff
@@ -65,8 +66,10 @@ class ChannelBase(BaseModel):
     private: bool = Field(True, title="channel should be private")
     size_limit: Optional[int] = Field(None, title="size limit of the channel")
     ttl: int = Field(36000, title="ttl of the channel")
-    mirror_channel_url: Optional[str] = Field(None, pattern="^(http|https)://.+")
-    mirror_mode: Optional[MirrorMode] = Field(None)
+    mirror_channel_url: Optional[Union[URLType, List[URLType]]] = Field(
```
AndreasAlbertQC, line 69: I feel like the goal of backward compatibility does not really warrant the confusion the `singular/plural` tighrope walk will induce here. Quetz is currently in major version `0`, we can afford to break backward compatibility in this very minor. Let's just go to a list and return a meaningful error in case someone tries to use the old way.

quetz/rest_models.py
```diff
@@ -65,8 +66,10 @@ class ChannelBase(BaseModel):
...
+        None,
+    )
+    mirror_mode: Optional[MirrorMode] = Field(None, nullable=True)
```
AndreasAlbertQC, line 72: Can we make this an enum so that only `proxy` and `mirror` are allowed?

quetz/rest_models.py
```diff
@@ -117,11 +120,22 @@ class ChannelActionEnum(str, Enum):
 
 
 class ChannelMetadata(BaseModel):
-    includelist: Optional[List[str]] = Field(
+
+    """
+    examples:
+    - includelist: ["numpy", "pandas"]
```
AndreasAlbertQC, line 126: Add a comment here explaining the difference

quetz/tasks/assertions.py
```diff
@@ -1,9 +1,9 @@
 def can_channel_synchronize(channel):
-    return channel.mirror_channel_url and (channel.mirror_mode == "mirror")
+    return channel.mirror_channel_urls and (channel.mirror_mode in ["mirror", "proxy"])
```
AndreasAlbertQC, line 2: this type of line is where I would really prefer mirror_mode to be an enum

quetz/tasks/mirror.py
```diff
@@ -197,6 +202,26 @@ def download_file(remote_repository, path_metadata):
     return f, package_name, metadata
 
 
+@retry(
+    stop=stop_after_attempt(3),
+    wait=wait_exponential(multiplier=1, min=4, max=10),
+    after=after_log(logger, logging.WARNING),
+)
+def _upload_package(file, channel_name, subdir, pkgstore):
```
AndreasAlbertQC, line 210: please add a docstring

quetz/tasks/mirror.py
```diff
@@ -235,38 +261,24 @@ def handle_repodata_package(
         total_size += size
         file.file.seek(0)
 
-    dao.assert_size_limits(channel_name, total_size)
+    # create package in database
```
AndreasAlbertQC, line 264: Should this be uncommented?

quetz/tasks/mirror.py
```diff
@@ -235,38 +261,24 @@ def handle_repodata_package(
...
     with TicToc("upload file without extracting"):
+        nthreads = config.general_package_unpack_threads
         with ThreadPoolExecutor(max_workers=nthreads) as executor:
             for file, package_name, metadata in files_metadata:
                 if proxylist and package_name in proxylist:
                     # skip packages that should only ever be proxied
                     continue
                 subdir = get_subdir_compat(metadata)
-                executor.submit(_upload_package, file, channel_name, subdir)
+                executor.submit(_upload_package, file, channel_name, subdir, pkgstore)
```
AndreasAlbertQC, line 281: Is there a behavior change here? Looks like you just moved the function def out of the current closure into the file scope. Anything else?

quetz/tasks/mirror.py
```diff
@@ -278,36 +290,22 @@ def _upload_package(file, channel_name, subdir):
...
-    use_repodata: bool = False,
-):
-    force = True  # needed for updating packages
-    logger.info(
-        f"Running channel mirroring {channel_name}/{arch} from {remote_repository.host}"
-    )
-
+def get_remote_repodata(
+    channel_name: str, arch: str, remote_repository: RemoteRepository
+) -> Union[dict, None]:
```
AndreasAlbertQC, line 295: Please add a docstring here that explains what the function does and most importantly how to interpret the return value (what does `None` mean?)

quetz/utils.py
```diff
@@ -15,28 +15,144 @@
...
+    channel_metadata: dict,
+    package_name: str,
+    package_metadata: dict,
+    remote_host: str,
+) -> MembershipAction:
+    """
+    Check if a package should be in a channel according
+    to the rules defined in the channel metadata.
+
+    Since for a given package and includelist, a package can be not in the includelist
```
AndreasAlbertQC, line 98:
```suggestion
    The function returns a representation of the treatment the package should receive (include / exclude / ignore). A package should be:
    * included if....
    * excluded if...
    * ignored if...
    
```

quetz/rest_models.py
AndreasAlbertQC, line 126: The `list` mode is just syntactic sugar so users with a single mirror channel don't need to type a dict, right? I feel like the extra complication might not be worth it in this case, and I would prefer to just have a single way of doing this rather than two every so slightly different ones. What do you thikn?

AndreasAlbertQC (changes requested):
@YYYasin19 thank you for this PR, great stuff! I reviewed ~half of the PR, feel free to already answer whatever makes sense. I hope to be able to pick up reviewing the rest of the PR later today.

quetz/tasks/mirror.py
```diff
@@ -317,10 +315,51 @@ def initial_sync_mirror(
...
             )
             if repodata_fn == "repodata.json":
-                return
+                return None
 
-    channel = dao.get_channel(channel_name)
+    return {}
+
+
+def initial_sync_mirror(
```
AndreasAlbertQC, line 323: Why is this function called `initial`?  Does it not cover syncs after the first one?

quetz/tasks/mirror.py
```diff
@@ -404,27 +444,46 @@ def handle_batch(update_batch):
...
+                metadata,
+                remote_host=remote_repository.host,
+            )
+            if action == MembershipAction.INCLUDE:
                 # try to find out whether it's a new package version
-
                 is_uptodate = None
                 for _check in version_checks:
-                    is_uptodate = _check(package_name, metadata)
+                    is_uptodate = _check(repo_package_name, metadata)
```
AndreasAlbertQC, line 461: Why `repo_package_name` instead of `package_name`?

quetz/tasks/mirror.py
```diff
@@ -434,10 +493,84 @@ def handle_batch(update_batch):
...
+    logger.debug(f"Removing {len(remove_batch)} packages: {remove_batch}")
+    removal_performed = False
+    package_specs_remove = set([p[1].split("-")[0] for p in remove_batch])
+    for package_specs in package_specs_remove:
+        dao.remove_package(channel.name, package_name=package_specs)
+        # TODO: is this needed every time?
+        dao.cleanup_channel_db(channel.name, package_name=package_specs)
+        if pkgstore.file_exists(channel.name, package_specs):
+            pkgstore.delete_file(channel.name, destination=package_specs)
+        removal_performed |= True
```
AndreasAlbertQC, line 525:
```suggestion
        removal_performed = True
```

quetz/tasks/mirror.py
```diff
@@ -540,37 +699,30 @@ def synchronize_packages(
...
         return
 
-    host = new_channel.mirror_channel_url
+    for mirror_channel_url in new_channel.mirror_channel_urls:
+        remote_repo = RemoteRepository(mirror_channel_url, session)
 
-    remote_repo = RemoteRepository(new_channel.mirror_channel_url, session)
-
-    user_id = auth.assert_user()
+        auth.assert_user()
```
AndreasAlbertQC, line 705: Why doe we not need the user ID anymore here?

AndreasAlbertQC (changes requested):
here's the second half -> let's meet up and talk some time

pyproject.toml
```diff
@@ -20,7 +20,7 @@ check-imports = ["quetz"]
 ignore = ["W004"]
 
 [tool.tbump.version]
-current = "0.9.2"
+current = "0.10.0"
```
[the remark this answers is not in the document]
YYYasin19: Yes, I just used it for deployment reasons on the dev environment so we can revert this before merging

quetz/db_models.py
```diff
@@ -200,14 +200,32 @@ class Channel(Base):
...
-    mirror_channel_url = Column(String)
+    mirror_channel_url = Column("mirror_channel_url", String)
     mirror_mode = Column(String)
     channel_metadata = Column(String, server_default='{}', nullable=False)
     timestamp_mirror_sync = Column(Integer, default=0)
     size = Column(BigInteger, default=0)
     size_limit = Column(BigInteger, default=None)
     ttl = Column(Integer, server_default=f'{60 * 60 * 10}', nullable=False)  # 10 hours
 
+    @property
```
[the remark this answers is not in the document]
YYYasin19: I didn't want to change the database type since it would also complicate the schema for such a small change.
Obviously we don't have any concrete numbers but I guess most users would only want to mirror one other channel instead of multiple ones.

Are you in favor of having a mapping to a list of mirror channel urls?

quetz/rest_models.py
YYYasin19: I could also use the list when I mirror multiple channels but still have _one_ list of packages I want to let through or exclude.
Also, it makes this stay backward-compatible 😬

quetz/tasks/mirror.py
YYYasin19: Yes, confused me as well :)

quetz/tasks/mirror.py
YYYasin19: I think the latter is a bit clearer about that it's something like `numpy-1.23.4-py39hefdcf20_0.tar.bz2` and not just `numpy`

quetz/dao.py
```diff
@@ -344,7 +361,7 @@ def cleanup_channel_db(
             Package.channel_name == channel_name
         )
         if package_name:
-            all_packages = all_packages.filter(
+            all_packages = all_packages.join(PackageVersion).filter(
```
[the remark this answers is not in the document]
YYYasin19: When you use an attribute of another table without specifying a join on that table, sqlalchemy will create a cross join. 

```sql
SELECT ... 
FROM packages, package_versions # should packages JOIN package_versions ON...
WHERE packages.channel_name = ? AND package_versions.package_name = ?
```

I can't remember the exact details, but this caused the wrong packages to be removed.

quetz/main.py
YYYasin19: Let's go with the route to directly disable this on channel creation time

quetz/dao.py
```diff
@@ -311,10 +311,12 @@ def create_channel(
                 "only ASCII characters should be used in channel name"
             )
 
+        # note: due to backwards compatibility, the rest model still calls it
+        # 'mirror_channel_url' but the db model calls it 'mirror_channel_urls'
         channel = Channel(
             name=data.name,
             description=data.description,
-            mirror_channel_url=data.mirror_channel_url,
+            mirror_channel_urls=data.mirror_channel_url,
```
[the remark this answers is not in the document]
YYYasin19: Nope, looks a bit complicated but it used the setter even when called over the constructor so this part
```python3
    @mirror_channel_urls.setter
    def mirror_channel_urls(self, value):
        print("yo im being set with ", value)
        if isinstance(value, str):
            self.mirror_channel_url = value
        elif isinstance(value, list):
            self.mirror_channel_url = ";".join(value)
        else:
            self.mirror_channel_url = None  # type: ignore
```
is actually run which just populates the mirror_channel_url again.

quetz/tasks/mirror.py
YYYasin19: Because it was needed for `create_packages_from_channeldata` (will comment on that below)

docs/source/using/mirroring.rst
```diff
@@ -61,6 +61,18 @@ Creating a mirror channel is similar to creating proxy channels except that you
...
+.. code:: json
+
+   {
+     "name": "mirror-channel",
+     "private": false,
+     "mirror_channel_url": ["https://conda.anaconda.org/btel", "https://conda.anaconda.org/conda-forge"],
+     "mirror_mode": "mirror"
+   }
+
+Note that setting multiple mirror channel urls will work for mirror channels only. Proxy channels can only mirror a single channel and will therefore use only the first url in the list.
```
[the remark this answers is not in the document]
YYYasin19: ✅

quetz/tasks/mirror.py
YYYasin19: The problem is that if we create packages already create packages from the channel metadata here (which seems to be some kind of.. shortcut?) than we have problems filtering them out later.

some solutions
1. call this code if no includelist/excludelist is set -> then we already sync everything
2. do the filtering based on includelist/excludelist here as well -> meh

quetz/tasks/mirror.py
YYYasin19: Nope, that should be it. And because of that, I had to pass the `pkgstore` as well.

quetz/tasks/mirror.py
```diff
@@ -434,10 +493,84 @@ def handle_batch(update_batch):
...
+        remove_batch: list of (arch, repo_package_name) tuples
+            e.g. [('linux-64', 'foo-1.0-0.tar.bz2'), ...]
+        channel: the channel to remove packages from
+        dao
+        pkgstore
+    Returns True if any removals were performed.
+    """
+    logger.debug(f"Removing {len(remove_batch)} packages: {remove_batch}")
+    removal_performed = False
+    package_specs_remove = set([p[1].split("-")[0] for p in remove_batch])
```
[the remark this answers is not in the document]
YYYasin19: It actually doesn't!
Are you aware of any tooling that might support this and prevent me from writing some weird unreadable regex that no one wants to touch again?

quetz/tasks/mirror.py
YYYasin19: I've implemented no. 1 for now

quetz/dao.py
```diff
@@ -332,6 +334,21 @@ def create_channel(
 
         return channel
 
+    def remove_package(self, channel_name: str, package_name: str):
```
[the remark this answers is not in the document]
YYYasin19: This is actually [already the case apparently](https://github.com/YYYasin19/quetz/blob/fc5c4bf5e839c776ec63209fad93640549034370/quetz/db_models.py#L365). I just missed it and checked for the forward refs. Thanks!

quetz/main.py
YYYasin19: It has been done

quetz/main.py
```diff
@@ -742,29 +742,32 @@ def post_channel(
...
-            json={
-                "url": api_endpoint.replace("api/channels", "get"),
-                "api_endpoint": api_endpoint,
-                "metrics_endpoint": api_endpoint.replace("api", "metrics"),
-            },
-            headers=headers,
-        )
-        if response.status_code != 201:
-            logger.warning(f"could not register mirror due to error {response.text}")
+        for mirror_url in channel.mirror_channel_urls:
```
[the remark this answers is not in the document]
YYYasin19: That's my understanding as well.

YYYasin19:
idea: We could, based on the "register yourself as a mirror on the source channel feature" also implement a trigger that sends a sync action to each "listening" mirror to update itself with the new package, i.e.: `mirror1` mirrors the channel `channel1`. We add a package to `channel1` which then sends a `synchronize_packages` action to all its mirrors.

quetz/tasks/assertions.py
YYYasin19: There already is an enum under `rest_models`. Done!

quetz/tasks/mirror.py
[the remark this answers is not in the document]
YYYasin19: Alright, I've used `Dist` from `conda`. 
-> We'll need to have a discussion about if we want to include conda as a dependency (which it already is in our dev environment) or vendor

quetz/rest_models.py
YYYasin19: Is this not already an enum..?

quetz/rest_models.py
```diff
@@ -7,12 +7,19 @@
...
+    ConfigDict,
+    Field,
+    constr,
+    field_validator,
+    model_validator,
+)
 
 T = TypeVar('T')
-URLType = Annotated[str, Field(pattern="^(http|https)://.+")]
+URLType = NewType('URLType', constr(pattern="^(http|https)://.+|None|^(\\[.*\\])+$"))
```
ivergara, line 22: Why do you have a URL type and define a pattern? https://docs.pydantic.dev/latest/usage/types/urls/

YYYasin19: Hm, my thought was that due to implementation of `AnyHttpUrl` in Pydantic
```
AnyHttpUrl = Annotated[Url, UrlConstraints(allowed_schemes=['http', 'https'])]
```
this won't work as well.
But afaik, they only require Python 3.7 which we do as well.

PR #632, opened 2023-05-31
enhancement: Make sqlalchemy pool settings configurable

AndreasAlbertQC:
This PR adds configuration settings for the sqlalchemy pool size and overflow settings.

Additionally, the docs for the sqlalchemy configuration section are extended. While doing that, I realized that there is a setting `database_plugin_path`, which afaict [is unused](https://github.com/search?q=repo%3Amamba-org%2Fquetz+database_plugin_path&type=code). I have not removed it for now, but if anyone knows what it's about, let me know.

PR #641, opened 2023-06-02
Do not pass pooling arguments to sqlite

### LINKED ISSUE #640: Rename postgres_pool_size and postgres_max_overflow
The postgres prefix is misplaced here:

https://github.com/mamba-org/quetz/blob/282f8ab1fd0d5a955815bf6246e461dc5b74fba8/quetz/database.py#L82-L83

AndreasAlbertQC:
* Pooling arguments in #632 were erroneously also propagated to sqlite engines. They should only be used if a postgres backend is used. This PR changes the propagation to fix that.
* I also reduced the default pool size from `100` to `10`, which should be much more appropriate for the typical quetz use cases, where there is no reason to maintain 100 open connections

Closes #640

### DIFF
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -30,10 +30,12 @@ Quetz can be run with SQLlite or PostgreSQL as database backends (PostgreSQL is 
 
    # The pool size for sqlalchemy engine connections to postgres DBs
    # see https://docs.sqlalchemy.org/en/latest/core/pooling.html
-   postgres_pool_size = 100
+   # Ignored for sqlite data bases
+   postgres_pool_size = 10
 
    # The maximal number of overflow connections beyond the pool size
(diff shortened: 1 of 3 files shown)


janjagusch (commented):
Thanks for the fix! The config still applies to database systems outside of Postgres, like MySQL, right? If so, I would welcome a naming change, as suggested in #640.

AndreasAlbertQC:
@janjagusch does quetz generally even claim to support MySQL? I was operating under the assumption that the data base is always either sqlite or postgres. As long as there are no docs or tests for MySQL, I would treat it as not supported.

ivergara:
In principle, one advantage of using an ORM like `sqlalchemy` is that you get database agnostic interface. So, unless quetz is doing weird stuff, it should be supported. However, I agree with @AndreasAlbertQC that without an explicit integration test is better not to openly claim MySQL support.

janjagusch (approved):

PR #643, opened 2023-06-08
Bump pyright-python version

rominf:
On my machine (openSUSE Tumbleweed) `pre-commit` creates dozens of
`npm` and `pyright` processes and linting never finishes.

The latest version of `pyright-python` does not have this issue.

PR #644, opened 2023-06-08
Use PEP-593 Annotated for options and arguments in CLI commands

rominf:
`typer` 0.9.0 introduced the support for PEP-593 `Annotated` for
specifying options and argument and it is now recommended to use it:
https://typer.tiangolo.com/release-notes/#090


rominf:
Depends on: #643. I will do a rebase after that PR is merged.

rominf:
@janjagusch I can't add label `maintenance` (most probably because of permissions), but GitHub Action requires some label. Could you please add it for me?

PR #647, opened 2023-06-09
Fix crash when uploading a package through scoped API key

### LINKED ISSUE #642: Unable to list package version after uploading via channel-scoped API token.
After uploading a package using a channel-scoped API token, listing the available package versions will fail with "500 Internal Server Error".

---

Steps to reproduce in the Quetz demo instance from the README:

1. log in to Quetz:
    ```console
    $ curl --cookie-jar "/tmp/cookies.txt" -s http://localhost:8000/api/dummylogin/alice`
    ```
2. create API key
    ```console
    $ curl -s -X POST http://localhost:8000/api/api-keys --cookie /tmp/cookies.txt --header 'Content-Type: application/json' --data '{ "description": "test token", "roles": [ { "channel": "test", "role": "maintainer" } ] }'
    {
      "description": "test token",
      "time_created": "2023-06-07",
      "expire_at": null,
      "roles": [
        {
          "channel": "test",
          "package": null,
          "role": "maintainer"
        }
      ],
      "key": "vDUuGngYoGKqRHJSSJWs1BUFtfE6T2jT"
    }
   ```
3. upload a package using the new token
	```console
    $ curl --header "X-API-Key: $QUETZ_API_KEY" --form "files=@xtensor/linux-64/xtensor-0.16.1-0.tar.bz2" http://localhost:8000/api/channels/test/files/
    null
	```
4. try to obtain its version information
	```console
    $ curl --get --header "X-API-Key: $QUETZ_API_KEY" http://localhost:8000/api/channels/test/packages/xtensor/versions
    Internal Server Error
	```

---

Stack trace in Quetz:

```
INFO:     127.0.0.1:59270 - "GET /api/channels/test/packages/xtensor/versions HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
[...]
  File "/home/user/miniconda3/envs/quetz/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/home/user/miniconda3/envs/quetz/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/home/user/quetz/quetz/main.py", line 1063, in get_package_versions
    version_data = rest_models.PackageVersion.from_orm(version)
  File "pydantic/main.py", line 579, in pydantic.main.BaseModel.from_orm
pydantic.error_wrappers.ValidationError: 1 validation error for PackageVersion
uploader
  none is not an allowed value (type=type_error.none.not_allowed)
```

---

I think what happens is, that it tries to look up the user information of the uploading "user".
But since the API key's `user_id` refers to a "virtual" user (with NULL value for its `username`database entry), it fails.

gabm:
fixes #642 

## Problem

- the uploader_id gets set to the user_id of the uploader when uploading package files
  - in case a scoped API key is used, that user_id points to an "anonymous user"
- when listing packages, that anonymous user does not have a name (null) and provokes a crash

## Solution

- when uploading through an API key, always retrieve the owner_id and use that as uploader_id

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -913,6 +913,8 @@ def post_package(
     dao: Dao = Depends(get_dao),
 ):
     user_id = auth.assert_user()
+    owner_id = auth.get_owner()
+
     auth.assert_create_package(channel.name)
     pm.hook.validate_new_package(
         channel_name=channel.name,
@@ -927,7 +929,7 @@ def post_package(
             detail=f"Package {channel.name}/{new_package.name} exists",
         )
 
-    dao.create_package(channel.name, new_package, user_id, authorization.OWNER)
+    dao.create_package(channel.name, new_package, owner_id, authorization.OWNER)
 
 
 @api_router.get(
@@ -1382,13 +1384,17 @@ async def post_upload(
             status_code=status.HTTP_406_NOT_ACCEPTABLE, detail="Wrong SHA256 checksum"
         )
 
-    user_id = auth.assert_user()
+    _ = auth.assert_user()
+
     auth.assert_create_package(channel_name)
     condainfo = CondaInfo((body), filename)
     dest = os.path.join(condainfo.info["subdir"], filename)
 
     body.seek(0)
     await pkgstore.add_package_async(body, channel_name, dest)
+
+    # get the id of the owner, in case auth was done through an API key
+    owner_id = auth.get_owner()
 
     package_name = str(condainfo.info.get("name"))
     package_data = rest_models.Package(
@@ -1400,7 +1406,7 @@ async def post_upload(
         dao.create_package(
             channel_name,
             package_data,
-            user_id,
+            owner_id,
             authorization.OWNER,
         )
 
@@ -1419,7 +1425,7 @@ async def post_upload(
             size=condainfo.info["size"],
             filename=filename,
             info=json.dumps(condainfo.info),
-            uploader_id=user_id,
+            uploader_id=owner_id,
             upsert=force,
         )
     except IntegrityError:
@@ -1504,7 +1510,8 @@ def handle_package_files(
     package=None,
     is_mirror_op=False,
 ):
-    user_id = auth.assert_user()
+    _ = auth.assert_user()
+    owner_id = auth.get_owner()
 
     # quick fail if not allowed to upload
     # note: we're checking later that `parts[0] == conda_info.package_name`
@@ -1648,7 +1655,7 @@ def handle_package_files(
             dao.create_package(
                 channel.name,
                 package_data,
-                user_id,
+                owner_id,
                 authorization.OWNER,
             )
 
@@ -1669,7 +1676,7 @@ def handle_package_files(
                 size=condainfo.info["size"],
                 filename=file.filename,
                 info=json.dumps(condainfo.info),
-                uploader_id=user_id,
+                uploader_id=owner_id,
                 upsert=force,
             )
         except IntegrityError:
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -912,8 +912,10 @@ def post_package(
     auth: authorization.Rules = Depends(get_rules),
     dao: Dao = Depends(get_dao),
 ):
-    user_id = auth.assert_user()
-    owner_id = auth.get_owner()
+    # here we use the owner_id as user_id. In case the authentication
+    # was done using an API Key, we want to attribute the uploaded package
+    # to the owner of that API Key and not the anonymous API Key itself.
+    user_id = auth.assert_owner()
 
     auth.assert_create_package(channel.name)
     pm.hook.validate_new_package(
@@ -929,7 +931,7 @@ def post_package(
             detail=f"Package {channel.name}/{new_package.name} exists",
         )
 
-    dao.create_package(channel.name, new_package, owner_id, authorization.OWNER)
+    dao.create_package(channel.name, new_package, user_id, authorization.OWNER)
 
 
 @api_router.get(
@@ -1384,7 +1386,10 @@ async def post_upload(
             status_code=status.HTTP_406_NOT_ACCEPTABLE, detail="Wrong SHA256 checksum"
         )
 
-    _ = auth.assert_user()
+    # here we use the owner_id as user_id. In case the authentication
+    # was done using an API Key, we want to attribute the uploaded package
+    # to the owner of that API Key and not the anonymous API Key itself.
+    user_id = auth.assert_owner()
 
     auth.assert_create_package(channel_name)
     condainfo = CondaInfo((body), filename)
@@ -1392,9 +1397,6 @@ async def post_upload(
 
     body.seek(0)
     await pkgstore.add_package_async(body, channel_name, dest)
-
-    # get the id of the owner, in case auth was done through an API key
-    owner_id = auth.get_owner()
 
     package_name = str(condainfo.info.get("name"))
     package_data = rest_models.Package(
@@ -1406,7 +1408,7 @@ async def post_upload(
         dao.create_package(
             channel_name,
             package_data,
-            owner_id,
+            user_id,
             authorization.OWNER,
         )
 
@@ -1425,7 +1427,7 @@ async def post_upload(
             size=condainfo.info["size"],
             filename=filename,
             info=json.dumps(condainfo.info),
-            uploader_id=owner_id,
+            uploader_id=user_id,
             upsert=force,
         )
     except IntegrityError:
@@ -1510,8 +1512,10 @@ def handle_package_files(
     package=None,
     is_mirror_op=False,
 ):
-    _ = auth.assert_user()
-    owner_id = auth.get_owner()
+    # here we use the owner_id as user_id. In case the authentication
+    # was done using an API Key, we want to attribute the uploaded package
+    # to the owner of that API Key and not the anonymous API Key itself.
+    user_id = auth.assert_owner()
 
     # quick fail if not allowed to upload
     # note: we're checking later that `parts[0] == conda_info.package_name`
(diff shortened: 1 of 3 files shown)


janjagusch:
@gabm, can you please rebase with the latest commit on main?

gabm:
done

janjagusch:
@gabm, can you fix the lint, please?

gabm:
actually the linter was very helpful.. I fixed the problems.. pls have a look

gabm:
I am actually not sure whether we also want to use the owner in other places... the theory goes like this

- the user might be a "virtual user" corresponding to "downgraded" privileges
- the owner is the "physical user" that issued the API key and therefore the "parent" of the "virtual user" above

In some places we want to use one and in other places the other. For example when uploading a package we want to set the owner as the owner of the pkgs, but in other places we need the reduced privileges of the virtual user... 

I am not familiar enough to review all the other uses or `get_user` and `assert_user`, but some might be replaced the owner method in some places. But then again its "dangerous" to touch, as it also escalates the privileges again..

quetz/main.py
```diff
@@ -912,7 +912,8 @@ def post_package(
     auth: authorization.Rules = Depends(get_rules),
     dao: Dao = Depends(get_dao),
 ):
-    user_id = auth.assert_user()
+    owner_id = auth.assert_owner()
```
AndreasAlbertQC, line 915: Since the argument to `create_package` still requires us to pass in a `user_id`, I'd recommend to still keep the line as `user_id = auth.assert_owner()` here, and add a comment that explains why we are using the user id of the owner here.

quetz/main.py
```diff
@@ -927,7 +928,7 @@ def post_package(
             detail=f"Package {channel.name}/{new_package.name} exists",
         )
 
-    dao.create_package(channel.name, new_package, user_id, authorization.OWNER)
+    dao.create_package(channel.name, new_package, owner_id, authorization.OWNER)
```
AndreasAlbertQC, line 931:
```suggestion
    dao.create_package(channel.name, new_package, user_id, authorization.OWNER)
```

quetz/main.py
```diff
@@ -1382,7 +1383,9 @@ async def post_upload(
             status_code=status.HTTP_406_NOT_ACCEPTABLE, detail="Wrong SHA256 checksum"
         )
 
-    user_id = auth.assert_user()
+    # get the id of the owner, in case auth was done through an API key
+    owner_id = auth.assert_owner()
```
AndreasAlbertQC, line 1387: Probably same comment here to keep it consistent: `user_id` instead of `owner_id` for consistency with `dao.create_package`

quetz/main.py
```diff
@@ -1504,7 +1507,7 @@ def handle_package_files(
     package=None,
     is_mirror_op=False,
 ):
-    user_id = auth.assert_user()
```
AndreasAlbertQC, line 1507: same same

AndreasAlbertQC (changes requested):
@gabm thanks for the PR. I appreciate the clarity of your fix. On the code itself, I mostly have nitpicky suggestions, I hope you don't mind. 
One actually important thing though: Would you mind also writing a test that captures the bug (==fails with the old implementation, but works with your fix?)

AndreasAlbertQC:
> I am actually not sure whether we also want to use the owner in other places... the theory goes like this
> 
> * the user might be a "virtual user" corresponding to "downgraded" privileges
> * the owner is the "physical user" that issued the API key and therefore the "parent" of the "virtual user" above
> 
> In some places we want to use one and in other places the other. For example when uploading a package we want to set the owner as the owner of the pkgs, but in other places we need the reduced privileges of the virtual user...
> 
> I am not familiar enough to review all the other uses or `get_user` and `assert_user`, but some might be replaced the owner method in some places. But then again its "dangerous" to touch, as it also escalates the privileges again..

My intuition here would be that the current implementation before your fix, where an anonymous user is used, is the most robust in the sense that it never leaks permissions. If anything, it is problematic because you sometimes get too few permissions. Therefore, unless we are fixing a specific problem (as you did in this PR), I would not change the existing system for fear of accidentally leaking permissions for no gain.

gabm:
thanks for those hints @AndreasAlbertQC . I added the tests as well.. pls have a look.

P.S.: I had a hard time working through the pytest fixtures, they are all over the place :/

quetz/tests/api/test_main_packages.py
```diff
@@ -732,3 +734,140 @@ def test_package_channel_data_attributes(
...
     assert content['platforms'] == ['linux-64']
     assert content['url'].startswith("https://")
+
+
+@pytest.fixture
+def owner(db, dao: Dao):
+    # create an owner with OWNER role
+    owner = dao.create_user_with_role("owner", role=SERVER_OWNER)
+
+    yield owner
```
AndreasAlbertQC, line 744:
```suggestion
try:
    yield owner
finally:
   db.delete(owner)
   db.commit()
```

quetz/tests/api/test_main_packages.py
```diff
@@ -732,3 +734,140 @@ def test_package_channel_data_attributes(
...
+    db.commit()
+
+
+@pytest.fixture
+def private_channel(db, dao: Dao, owner):
+    # create a channel
+    channel_data = Channel(name='private-channel', private=True)
+    channel = dao.create_channel(channel_data, owner.id, "owner")
+
+    yield channel
```
AndreasAlbertQC, line 756: `try...finally`

quetz/tests/api/test_main_packages.py
```diff
@@ -732,3 +734,140 @@ def test_package_channel_data_attributes(
...
+    db.commit()
+
+
+@pytest.fixture
+def private_package(db, owner, private_channel, dao):
+    package_data = Package(name='test-package')
+
+    package = dao.create_package(private_channel.name, package_data, owner.id, "owner")
+
+    yield package
```
AndreasAlbertQC, line 768: `try...finally`

quetz/tests/api/test_main_packages.py
```diff
@@ -732,3 +734,140 @@ def test_package_channel_data_attributes(
...
+                        'package': None,
+                        'channel': private_channel.name,
+                    }
+                ],
+            )
+        ),
+        "API key with role restruction",
+    )
+
+    yield key
```
AndreasAlbertQC, line 795: `try...finally`

AndreasAlbertQC (changes requested):
Thanks for the additional work and tests @gabm ! Tests look good to me, I would just add a a try/finally around the yields in the fixtures to headaches if a test throws an exception. Apart from that, this is ready to go IMO

beenje:
Hi! Thanks for working on this. I have the same issue :-)

I did a quick test and this branch indeed works for new uploaded packages.
What about existing packages? Should this PR include a migration script to fix the database?

Also just wondering if an alternative solution could not have been to show "anonymous/none" for the uploader.
There must be a reason for the user_id to be set to this user?

AndreasAlbertQC (approved):
As @ivergara kindly pointed out to me, pytest already takes care of the exception handling for yielding fixtures such that the try/except block I asked for is not needed. My bad!

AndreasAlbertQC:
The CI is fixed on `main` (pydantic dependency pin). @gabm rebasing your branch on `main` should fix the CI here.

gabm:
done

gabm:
> What about existing packages? Should this PR include a migration script to fix the database?

For us its not a huge problem, we fixed it manually so far. Currently there is no straight forward way to fix it in the DB because the uploader information is lost iirc.

> Also just wondering if an alternative solution could not have been to show "anonymous/none" for the uploader. There must be a reason for the user_id to be set to this user?

The full details are outlined here #642. In short: the current user model mixes "identity" and "authorization" to some extent. When you create an API key with a scope, then a new user without username is created and added as a member/maintainer to the packages/channels etc... When uploading a package with that API Key, it correctly resolves to that new, anonymous user that has the restricted authorization - thus preventing privilege escalation. Unfortunately, as this anonymous user has no identity (no name etc), the uploader field remains unset and that causes the error. The uploader field should be set with the identity of the "owner of the API Key". This identity should imho not be "anonymous"...

beenje:
> > Also just wondering if an alternative solution could not have been to show "anonymous/none" for the uploader. There must be a reason for the user_id to be set to this user?
> 
> The full details are outlined here #642. In short: the current user model mixes "identity" and "authorization" to some extent. When you create an API key with a scope, then a new user without username is created and added as a member/maintainer to the packages/channels etc... When uploading a package with that API Key, it correctly resolves to that new, anonymous user that has the restricted authorization - thus preventing privilege escalation. Unfortunately, as this anonymous user has no identity (no name etc), the uploader field remains unset and that causes the error. The uploader field should be set with the identity of the "owner of the API Key". This identity should imho not be "anonymous"...

Thanks for the additional information about restricted authorization. I agree using the owner_id makes sense.

> > What about existing packages? Should this PR include a migration script to fix the database?
> 
> For us its not a huge problem, we fixed it manually so far. Currently there is no straight forward way to fix it in the DB because the uploader information is lost iirc.

I think it's possible to have a migration script:
- we can get null users from `users` table
- in the `api_keys` table, we have the link user_id / owner_id
- In `package_members` table, we want to replace the user_id of null users with owner_id from the `api_keys`

AndreasAlbertQC:
Thanks @gabm!

@beenje Do you want to break out the migration followup into a separate issue or PR? In the interest of somewhat timely turnaround, I'd like to merge this PR now so that we at least have the fix in the `main`.

beenje:
> Thanks @gabm!
> 
> @beenje Do you want to break out the migration followup into a separate issue or PR? In the interest of somewhat timely turnaround, I'd like to merge this PR now so that we at least have the fix in the `main`.

Yes, go ahead! I will try to create another PR for the migration.

PR #648, opened 2023-06-10
Use Pydantic as Config backend

janjagusch:
As first attempt in using Pydantic as a config backend.


quetz/tests/test_config.py
```diff
@@ -35,26 +35,29 @@ def test_config_without_file_path_set(config_str):
...
-    assert not config.users_admins
-    assert not config.users_maintainers
-    assert not config.users_members
-    assert not config.users_default_role
-    assert not config.users_create_default_channel
+    assert not config.users.admins
+    assert not config.users.maintainers
+    assert not config.users.members
+    assert not config.users.default_role
+    assert not config.users.create_default_channel
```
janjagusch, line 56: Accessing fields through the `_` operator will not be possible anymore.

quetz/config.py
```diff
@@ -24,207 +25,276 @@
...
+        cls.__annotations__.update(new_annotations)
+
+
+class SettingsGeneral(BaseSettings):
+    package_unpack_threads: int = 1
+    frontend_dir: str = ""
+    redirect_http_to_https: bool = False
+
+    class Config:
+        env_prefix = 'general'
```
ivergara, line 100: You probably will need to add a `_`suffix to all the `env_prefix` if not you'll have something like `generalredirect_http_to_https` environment variable which I don't think it's what you want.

quetz/config.py
```diff
@@ -24,207 +25,276 @@
...
+class SettingsGitHub(BaseSettings):
+    client_id: str
+    client_secret: str
+
+    class Config:
+        env_prefix = 'github'
+
+
+class SettingsGitLab(BaseSettings):
+    url: str = "https://gitlab.com"
```
ivergara, line 122: You can use a `pydantic` [type](https://docs.pydantic.dev/latest/usage/types) to validate [urls](https://docs.pydantic.dev/latest/usage/types/#urls).

quetz/config.py
```diff
@@ -24,207 +25,276 @@
...
+    package_unpack_threads: int = 1
+    frontend_dir: str = ""
+    redirect_http_to_https: bool = False
+
+    class Config:
+        env_prefix = 'general'
+
+
+class SettingsCORS(BaseSettings):
+    allow_origins: list = []
```
ivergara, line 104:
```suggestion
    allow_origins: list[str] = []
```
I guess these aren't just strings but urls, see another comment on that (in the gitlab settings)

quetz/config.py
```diff
@@ -24,207 +25,276 @@
...
+    class Config:
+        env_prefix = 'users'
 
-class ConfigSection(NamedTuple):
-    name: str
-    entries: List[ConfigEntry]
-    required: bool = True
+
+class SettingsWorker:
+    type: str = "thread"
```
ivergara, line 232: I don't know which options there are for the type of worker, but I guess it's limited. Perhaps use a `Literal` here with the options? This is rather minor and not critical for this PR.

quetz/config.py
```diff
@@ -24,207 +25,276 @@
...
+
+class SettingsProfiling(BaseSettings):
+    enable_sampling: bool = False
+    interval_seconds: float = 0.001
+
+    class Config:
+        env_prefix = 'profiling'
+
+
+class Settings(DyanamicallyExtendableSetting):
```
ivergara, line 272: What about sorting all the required settings first and then the optional ones?

ivergara:
```suggestion
class QuetzSettings(DyanamicallyExtendableSetting):
```

quetz/tests/test_config.py
```diff
@@ -83,56 +86,61 @@ def test_config_with_path(config_dir, config_base):
...
+def _setup_config_extend_require():
+    from pydantic import BaseSettings
+
+    class OtherPluginSetting(BaseSettings):
+        some_config_value: str
+
+    Config.register(other_plugin=(OtherPluginSetting, ...))
+    yield
+    del Settings.__fields__["other_plugin"]
+    del Settings.__annotations__["other_plugin"]
```
ivergara, line 111: Do you need to explicitly do cleanup?

Can't you control this via the scope of the fixture?

quetz/tests/test_config.py
ivergara: I'd say good riddance!

quetz/config.py
janjagusch: I'm also not sure. In the first pass, I was just copying from the previous config backend. Once this is all working I will convert this to an enum.

quetz/tests/test_config.py
janjagusch: Might be difficult, since we're changing a class, not an instance of a class.

PR #649, opened 2023-06-15
Add performance-oriented dependencies

ivergara:
`uvicorn` can use performance oriented implementations of the event loop (uvloop) and http parsing (httptools) if they're present. This Pr adds those into the dependencies.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,6 @@
 name: quetz
-channels:
-  - https://repo.mamba.pm/conda-forge
+ channels:
+   - https://repo.mamba.pm/conda-forge
 dependencies:
   - python>=3.7
   - pip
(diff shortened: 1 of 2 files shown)


janjagusch:
@ivergara, can you please rebase with the latest commit on main?

janjagusch:
@ivergara, the diff still looks off. Can you try merging the latest commit on main into your branch, please?

ivergara:
Sorry, I did somehow botch the rebase.

janjagusch (changes requested):
I'm unsure if we want the two packages to be part of `install_requires`. This would imply that you must have these packages installed. There might be some use cases where this is not possible. I'm fine with adding it to the `environment.yml`, though.

What do you think?

ivergara:
Adding it to the `environment.yml` makes little sense in my opinion. That'd only affect a local development deployment.

I do agree that having them by default might be controversial. I deemed it as a safe operation. One place it'd blow up is isntalling quetz in windows.

I'm fine with closing this.

PR #650, opened 2023-06-16
Fix event loop errors

janjagusch:


quetz/pkgstores.py
```diff
@@ -302,8 +303,12 @@ def __init__(self, config):
         key = config['key'] if config['key'] != '' else None
         secret = config['secret'] if config['secret'] != '' else None
         self.fs = s3fs.S3FileSystem(
-            key=key, secret=secret, asynchronous=True, client_kwargs=client_kwargs
+            key=key,
+            secret=secret,
+            asynchronous=True,
+            client_kwargs=client_kwargs,
         )
+        self.fs._loop = get_loop()
```
janjagusch, line 311: @martinRenou, I hope you don't mind that I'm looping you in here. This is a follow-up from our conversation in https://github.com/fsspec/filesystem_spec/pull/1290. You suggested to explicitly pass the `loop` argument to the file system constructor. However, this doesn't seem possible, since whenever `asynchronous=True` then `self._loop=None` (see [here](https://github.com/fsspec/filesystem_spec/blob/960866a7d5e780000924380d75cadbdc66faa24b/fsspec/asyn.py#L308-L311)).

The only solution I could come up with is to set the `self._loop` attribute after the object has been instantiated. Is this a bug in `fsspec` or am I doing something wrong?

Thanks a lot for your help!

janjagusch: Ah, sorry about that, @martinRenou. :)

janjagusch, line 311: @martindurant, I hope you don't mind that I'm looping you in here. This is a follow-up from our conversation in https://github.com/fsspec/filesystem_spec/pull/1290. You suggested to explicitly pass the loop argument to the file system constructor. However, this doesn't seem possible, since whenever asynchronous=True then self._loop=None (see [here](https://github.com/fsspec/filesystem_spec/blob/960866a7d5e780000924380d75cadbdc66faa24b/fsspec/asyn.py#L308-L311)).

The only solution I could come up with is to set the self._loop attribute after the object has been instantiated. Is this a bug in fsspec or am I doing something wrong?

Thanks a lot for your help!

martindurant: Perhaps the easiest is to follow the pattern in a test: https://github.com/fsspec/s3fs/blob/main/s3fs/tests/test_s3fs.py#L2000

I am sure this is also documented, but where to look between s3fs's and fsspec's docs isn't always clear. Help always appreciated...

martindurant: but again, the best thing you can do is actually create the filesystem object within a coroutine, in which case the asynchronous and loop kwargs should not be necessary (but you cannot call any sync-API methods)

PR #653, opened 2023-07-20
Pin pydantic<2

AndreasAlbertQC:
Two issues:
1. pydantic was not in our environment.yml even though we are using it directly (not just as a transitive dependency)
2. our code is currently not compatible with the recently released pydantic 2.

This PR hotfixes the dependency by down pinning it. As a follow up, we should migrate our code to be compatible with pydantic 2 in the future.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -50,5 +50,6 @@ dependencies:
   - conda-content-trust
   - pyinstrument
   - pytest-asyncio
+  - pydantic <2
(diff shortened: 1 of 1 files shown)


AndreasAlbertQC:
Will need to port this change also to https://github.com/conda-forge/quetz-feedstock

janjagusch (approved):

PR #655, opened 2023-08-02
Add migration script for scoped API keys

beenje:
Add script to update `package_versions` uploader with API keys owner instead of the key anonymous user.

This is linked to https://github.com/mamba-org/quetz/issues/642 fixed in https://github.com/mamba-org/quetz/pull/647.
The script allows users to update existing databases automatically.


AndreasAlbertQC (approved):
Thanks @beenje ! Looks good to me

PR #656, opened 2023-08-05
Migrate to Pydantic v2

### LINKED ISSUE #654: upgrade to pydantic v2
As mentioned in #653 we should update to pydantic2 some time.
Aside from the [migration guide](https://docs.pydantic.dev/latest/migration/), there's also a [bump-pydantic](https://github.com/pydantic/bump-pydantic) tool that aims to do a lot of / all the things automatically!

beenje:
Migrate to Pydantic v2:

Had to force some fields to be `Optional` if they can be set to None.

- Ran [bump-pydantic](https://github.com/pydantic/bump-pydantic)
- Replace deprecated `parse_obj` and `from_orm` with `model_validate`
- Replace deprecated `dict` with `model_dump`
- Replace `update_forward_refs` with `model_rebuild`
- Remove `nullable=True` in pydantic Field. This isn't used. To mark a field as nullable, it should be set as Optional or "| None". Fix deprecation warning: Extra keyword arguments on `Field` is deprecated and will be removed.
- Add UserOptionalRole class. User role is nullable and should be optional in get (but not in set)
- Fix Job rest model. item_spec can't be set to required in JobBase and optional in Job (mypy complains). Create a new class JobCreate where item_spec is required. In job, item_spec is optional.
- Fix Packages and ChannelSearch rest models:
    - platforms is nullable: it should be optional
    - Channel description is optional - so should be ChannelSearch description

Fix #654

### DIFF
--- a/quetz/tests/test_jobs.py
+++ b/quetz/tests/test_jobs.py
@@ -607,7 +607,7 @@ def test_post_new_job_invalid_items_spec(auth_client, user, db, dummy_job_plugin
     )
     assert response.status_code == 422
     msg = response.json()['detail']
-    assert "not an allowed value" in msg[0]['msg']
+    assert "Input should be a valid string" in msg[0]['msg']
 
 
 @pytest.mark.parametrize("user_role", ["owner"])
--- a/quetz/tests/test_jobs.py
+++ b/quetz/tests/test_jobs.py
@@ -598,10 +598,14 @@ def test_post_new_job_manifest_validation(
 
 
 @pytest.mark.parametrize("user_role", ["owner"])
-def test_post_new_job_invalid_items_spec(auth_client, user, db, dummy_job_plugin):
+def test_post_new_job_invalid_items_spec(
+    auth_client, user, db, dummy_job_plugin, mocker
+):
     # items_spec=None is not allowed for jobs
     # (but it works with actions)
     manifest = "quetz-dummyplugin:dummy_func"
+    dummy_func = mocker.Mock()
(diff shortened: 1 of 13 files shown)


wolfv:
Thanks @beenje this looks really good to me! Awesome work. I would like to hear from QuantCo if they are happy with these changes (since they've been very active on recent PR's) but let's get this in! :)

quetz/tests/test_jobs.py
```diff
@@ -598,16 +598,20 @@ def test_post_new_job_manifest_validation(
...
 
 @pytest.mark.parametrize("user_role", ["owner"])
-def test_post_new_job_invalid_items_spec(auth_client, user, db, dummy_job_plugin):
+def test_post_new_job_invalid_items_spec(
+    auth_client, user, db, dummy_job_plugin, mocker
+):
     # items_spec=None is not allowed for jobs
     # (but it works with actions)
     manifest = "quetz-dummyplugin:dummy_func"
+    dummy_func = mocker.Mock()
```
ivergara, line 607: Why did you have to add this?

ivergara (commented):
After completing a migration myself, this looks good.

quetz/tests/test_jobs.py
beenje: Without this, the error raised is:

```
>       assert "Input should be a valid string" in msg[0]['msg']
E       assert 'Input should be a valid string' in "Value error, invalid function 'dummy_func' name in plugin 'quetz-dummyplugin'"
```

I guess the validation order changed in pydantic. The "manifest" value is checked before "items_spec".

ivergara (approved):

Issue #646, opened 2023-06-08
Bump mypy version to 1.3.0


rominf:
The latest `mypy` catches more bugs and is three times faster on my PC when running on the whole code base without caches than the current version (9 seconds vs. 27 seconds).

I propose bumping `mypy` version (and fixing all the issues, of course).

- [x] I am ready to create a PR.

wolfv:
This would be great @rominf – please go for it :)

rominf:
Hi @wolfv. I've done some preparation work in #658. Please have a look :-)

PR #658, opened 2023-08-30
mypy: analyze types in all possible libraries

rominf:
Preparation for #646

* Remove `ignore_missing_imports = true` from mypy configuration.
* Run mypy in the same environment instead of separate to check types
  in dependencies like fastapi.
* Move mypy dependencies from pre-commit configuration to `setup.cfg`.
* Update mypy dependencies there.
* Move `rq` from `environment.yml` to `setup.cfg`: conda-forge version:
  1.9.0, pypi version : 1.15.1 (two years difference; types were added).
* Add libraries with missing types to ignore list in mypy
  configuration.
* Add pydantic mypy plugin.
* Allow running mypy without explicit paths.
* Update GitHub Actions.
* Temporarily add ignore `annotation-unchecked` to make mypy pass.
* Fix new mypy issues:
  * Use https://github.com/hauntsaninja/no_implicit_optional to make
    `Optional` explicit.
  * If there is no default, the first `pydantic.Field` argument should
    be omitted (`None` means that the default argument is `None`).
  * Refactor `_run_migrations`. There were two different paths: one for
    normal execution and another one for testing. Simplify arguments and
    the function code, and introduce a new mock `run_migrations`.
  * To preserve compatibility, introduce `ChannelWithOptionalName` for
    the `/channels` patch method. Note that the solution is a bit dirty
    (I had to use `type: ignore[assignment]`) to minimize the number of
    models and the diff.
  * Trivial errors.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,6 +17,9 @@ jobs:
       - name: Add micromamba to GITHUB_PATH
         run: echo "${HOME}/micromamba-bin" >> "$GITHUB_PATH"
       - run: ln -s "${CONDA_PREFIX}" .venv # Necessary for pyright.
+      - run: pip install -e .[mypy]
+      - name: Add mypy to GITHUB_PATH
+        run: echo "${GITHUB_WORKSPACE}/.venv/bin" >> "$GITHUB_PATH"
       - uses: pre-commit/action@v3.0.0
         with:
           extra_args: --all-files --show-diff-on-failure
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,19 +19,7 @@ repos:
     hooks:
       - id: mypy
         files: ^quetz/
-        additional_dependencies:
-          - sqlalchemy-stubs
-          - types-click
-          - types-Jinja2
-          - types-mock
-          - types-orjson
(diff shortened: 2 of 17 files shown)


rominf:
> If there is no default, first `pydantic.Field` argument should be omitted (`None` means that the default argument is `None`).

but

> Introduce `ChannelWithOptionalName` for `/channels` patch method to preserve compatibility.

I'm relying on the test suite and/or your knowledge of what's `Optional` and what's not since the codebase is new to me and I might miss something.

rominf:
It looks like the test `test_download_remote_file_in_parallel` is flaky: there is a problem in the test or the code. I haven't touched this place, and the previous run with the same code succeeded (I just fixed the commit message in the last force push).

The link to the job: https://github.com/mamba-org/quetz/actions/runs/6065939126/job/16455972862?pr=658

wolfv:
Hey @rominf indeed, that test is unfortunately flaky (has been this way since a rather long time).

rominf:
Rebased, now tests pass.

rominf:
A reminder about the PR.

Feel free to close the PR if its continuation is not beneficial. I am open to rebasing and improving it though.

rominf:
Closing as stale. Feel free to reopen.

PR #662, opened 2023-09-07
add pam config section to config.py

mbestipa:
I'm trying to get to a point where I can deploy quetz without using a config file. I'm currently trying to use the pam authenticator to do so, but ran into issues because the way quetz loads the config from the environment only works when the authenticator is known to the `Config.config_map`. This should fix that


wolfv:
Hmm, is it not possible to find another fix for this so that we register the config earlier or something like that?

PR #663, opened 2023-09-08
Fix double upload bug and improve test

AndreasAlbertQC:
In our deployment we have been encountering a bug that happens when:
1. a package file is uploaded through the `/upload/` route.
2. a package file with the same package version inside and the same file name, but different hashes is uploaded again.

We observe that in the second upload, we get a 409, which is good. However, we also find that the binary package file is still updated on the server, even though the http request was rejected. The repodata is not updated, as expected. This means that we get hash mismatch errors because the incorrectly updated binary does not agree with the correct hashes for the original file in the repodata.

I think the problem is [here](https://github.com/mamba-org/quetz/blob/main/quetz/main.py#L1399), where we call `pkgstore.add_package_async` before the try/except block that ultimately results in the 409.

During debugging I found that the other upload route is also affected.

Changes in this PR:
* Extended the existing duplicate upload test:
   *  we now use files that have the same name. This is crucial to trigger the bug. In previous testing, files with different names were used.
   * we now also check the file that is actually present in the store. Previously, we just uploaded and checked the repodata, which leads us to miss the bug that manifests only in the stored files.
* Bug fix:
   * in `/channels/{channel_name}/packages/{package_name}/files/`, the upload is handled through a call to `handle_package_files`, which in turn uses `_extract_and_upload_package`. I added a `force` argument and a file existence check to  `_extract_and_upload_package`, which is then transformed into an HTTP 409 in `handle_package_files`
   * In `/channels/{channel_name}/upload/{filename}`, the solution is more straightforward as the call to `pkgstore.add_package_async` happens right in the top-level function. I moved the call after the try/except block that checks for 409.

Expected test failure before fix [here](https://github.com/mamba-org/quetz/actions/runs/6120084506/job/16611255172?pr=663)

### DIFF
--- a/quetz/main.py
+++ b/quetz/main.py
@@ -1395,9 +1395,6 @@ async def post_upload(
 
     dest = os.path.join(condainfo.info["subdir"], filename)
 
-    body.seek(0)
-    await pkgstore.add_package_async(body, channel_name, dest)
-
     package_name = str(condainfo.info.get("name"))
     package_data = rest_models.Package(
         name=package_name,
@@ -1433,6 +1430,9 @@ async def post_upload(
     except IntegrityError:
         logger.debug(f"duplicate package '{package_name}' in channel '{channel_name}'")
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail="Duplicate")
+
+    body.seek(0)
+    await pkgstore.add_package_async(body, channel_name, dest)
 
     pm.hook.post_add_package_version(version=version, condainfo=condainfo)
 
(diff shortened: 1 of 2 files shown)


quetz/main.py
```diff
@@ -1434,6 +1431,9 @@ async def post_upload(
         logger.debug(f"duplicate package '{package_name}' in channel '{channel_name}'")
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail="Duplicate")
 
+    body.seek(0)
```
AndreasAlbertQC, line 1434: here the bug fix is simply moving the call down

wolfv (approved):

ivergara (approved):

PR #670, opened 2023-11-13
Add HEAD authorization tests

jonashaag:
I'd like to make sure that `HEAD` requests respond with 40x iff the corresponding `GET` requests respond with 40x.

I didn't find any tests for the actual package URLs. I'd like to add tests for those URLs as well.

### DIFF
--- a/quetz/tests/test_auth.py
+++ b/quetz/tests/test_auth.py
@@ -439,20 +439,23 @@ def test_private_channels_download(db, client, data, channel_dirs):
     assert response.text == "file content 0"
 
     # fail on private channel without credentials
-    response = client.get('/get/privatechannel/noarch/current_repodata.json')
-    assert response.status_code == 401
+    for method in [client.get, client.head]:
+        response = method('/get/privatechannel/noarch/current_repodata.json')
+        assert response.status_code == 401
 
     # fail on private channel with invalid credentials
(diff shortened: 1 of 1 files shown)


quetz/tests/test_auth.py
```diff
@@ -439,20 +439,23 @@ def test_private_channels_download(db, client, data, channel_dirs):
     assert response.text == "file content 0"
 
     # fail on private channel without credentials
-    response = client.get('/get/privatechannel/noarch/current_repodata.json')
-    assert response.status_code == 401
+    for method in [client.get, client.head]:
```
ivergara, line 442: @jonashaag, what would speak against adding a method callable as a parametrization instead of adding an iteration on each assertion?

ivergara (approved):

quetz/tests/test_auth.py
jonashaag: That was my first approach. The problem is that you need this in very few places only

ivergara: Now I see it. 🙈

wolfv:
Looks good to me, but I assume we're waiting for some more tests @jonashaag ?

jonashaag:
Are there any tests for package downloads? I haven't found them. I'd like to add the same there

PR #671, opened 2023-11-21
ignore missing `info/files` file

wolfv:
cc @beenje do you think you could deploy a change like this?

### DIFF
--- a/quetz/condainfo.py
+++ b/quetz/condainfo.py
@@ -162,7 +162,7 @@ class CondaInfo:
         try:
             with tar.extractfile("info/files") as fp:
                 self.files = fp.readlines()
-        except:
+        except KeyError:
             self.files = [p["_path"] for p in self.paths.get("paths", [])]
 
         try:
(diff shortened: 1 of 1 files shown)


quetz/condainfo.py
```diff
@@ -158,8 +158,12 @@ def _load_jsons(self, tar):
             self.paths = json.load(tar.extractfile("info/paths.json"))
         except KeyError:
             self.paths = {}
-        with tar.extractfile("info/files") as fp:
-            self.files = fp.readlines()
+
+        try:
+            with tar.extractfile("info/files") as fp:
+                self.files = fp.readlines()
+        except:
```
beenje, line 165: Suggestion:

```suggestion
        except KeyError:
```

beenje (commented):
I'll try to test that locally!

beenje:
Tested that change locally. I uploaded a package created by rattler-build without issue. Could install it with mamba afterwards.

Looks good to me. Thanks!

pavelzw (approved):

pavelzw:
Available on dockerhub now 
image

PR #673, opened 2023-11-22
fix: Address issues of missing `bucket_name` in s3fs paths

RobinHolzingerQC:
## Motivation

The previous way of specifying paths for `s3fs` (not including the bucket_name) led to using when using `endpoint_url`.

## Changes

As paths in the format of `<root_dir>/<subdir>/.../<file>` are not officially supported by `s3fs` (see https://github.com/fsspec/s3fs/issues/824), we are now using `<bucket_name>/<root_dir>/<subdir>/.../<file>`.
For that, we need a `bucket_name` config field.
`def _bucket_map(...)` is adjusted according so that every channel specification also includes the `bucket_name`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ test_quetz
 .env
 .envrc
 .venv
+
+# OS
+.DS_Store
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
+## 0.10.2
+
+([Full Changelog](https://github.com/mamba-org/quetz/compare/v0.10.1...3cbd27a2356c9ff07fc80fef5888acb71fc2d7f7))
+
+### Bugs fixed
+
+- Fix set user roles when role is None [#669](https://github.com/mamba-org/quetz/pull/669) ([@janjagusch](https://github.com/janjagusch))
+
+### Contributors to this release
+
+([GitHub contributors page for this release](https://github.com/mamba-org/quetz/graphs/contributors?from=2023-09-28&to=2023-09-28&type=c))
+
+[@janjagusch](https://github.com/search?q=repo%3Amamba-org%2Fquetz+involves%3Ajanjagusch+updated%3A2023-09-28..2023-09-28&type=Issues)
+
+<!-- <END NEW CHANGELOG ENTRY> -->
+
 ## 0.10.1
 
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -96,6 +99,10 @@ jobs:
           fi
 
           export QUETZ_IS_TEST=1
+
+          echo "S3_BUCKET_NAME: $S3_BUCKET_NAME"
+          echo "QUETZ_S3_BUCKET_NAME: $QUETZ_S3_BUCKET_NAME"
```
janjagusch, line 104: I guess this was for debugging purposes and can be removed, right?

```suggestion
```

quetz/tests/test_pkg_stores.py
```diff
@@ -156,6 +156,9 @@ def test_remove_dirs(any_store, channel_name):
 
 @pytest.fixture
 def s3_store():
+    print(">>> s3_config", s3_config)
+    print(">>> s3_config.bucket_name", s3_config.get("bucket_name"))
+    print(">>> s3_config | os.environ", os.environ)
```
janjagusch, line 161:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -85,6 +86,8 @@ jobs:
           S3_SECRET_KEY: ${{ secrets.s3_secret_key }}
           S3_ENDPOINT: https://s3.sbg.cloud.ovh.net/
           S3_REGION: sbg
+          S3_BUCKET_NAME: quetz
+          QUETZ_S3_BUCKET_NAME: quetz
```
janjagusch, line 90: Can you brielfy elaborate what's the purpose of this env var?

RobinHolzingerQC: As I found out with https://github.com/fsspec/s3fs/issues/824, we need to specify the bucket_name in the paths for s3fs, we need to know how the bucket is called that we write to. Previously (at least for aws s3) this information could be encoded in `S3_ENDPOINT`, however, f3fs doesn't work with that (bucket cannot be encoded in URL). Therefore we need another variable explicitly stating the bucket_name.

janjagusch: But why do we need `S3_BUCKET_NAME` and `QUETZ_S3_BUCKET_NAME`?

RobinHolzingerQC: I already removed `QUETZ_S3_BUCKET_NAME` again, I just tested the `QUETZ_` prefix to make sure that it wouldn't be mapped to `S3_BUCKET_NAME` in the tests (as the env variable was missing)

PR #675, opened 2023-11-30
Add support for repodata.json.zst

### LINKED ISSUE #573: Add repodata.json.zst
Conda recently added a zstd compression version of repodata.json: https://github.com/conda/infra/issues/648

Would be great to have it here as well!

beenje:
Adding support for `repodata.json.zst` (fix #573).

- `repodata.json.zst` is now created as well as the `.bz2` and `.gz` versions
- added a test checking that the compressed files are identical to `repodata.json`
- Ensure that repodata.json is updated for proxy channels when a compressed file is requested. Before that fix, if  `repodata.json.zst` was requested and existed locally, it was never updated. Even with the support added in first commit, it would only be updated when someone was requesting the `repodata.json` file (which triggers the creation of all compressed files). Note that I didn't find an easy way to write a test for that, which would be nice.


beenje:
With the tests added in https://github.com/mamba-org/quetz/pull/677 I could easily add a test in this MR.

beenje:
I don't think async will help in this case. The compression is done in `add_static_file` (and `add_temp_static_file`), which aren't async. They are called in the background by `update_indexes`. There is no issue there.
But for proxy channel, we download the remote `repodata.json`, compress it and then serve it. The client has to wait during that time. Would be the same if it was async.
Compression should probably be done in the background for proxy channels as well.

In the meantime, I added a new compression section in the config to enable/disabled bz2, gz and zst compressions.
By default zst is disabled and the 2 others are enabled to keep the same behaviour as today.

Note that for the tests, I re-used what I implemented in https://github.com/mamba-org/quetz/pull/677. So I rebased this PR on the other one branch. Hoping #677 can be merged soon.

PR #687, opened 2024-02-11
feat: Support pre-configured supertoken authentication

RobinHolzingerQC-OLD:
# Motivation

We want to run automated CI tests on a dockerized version of quetz (deployed with helm/Kubernetes). To be able to use the API there without any real user (e.g. authenticated via GitHub) we would like to be able to define a supertoken in the TOML config. It should act as an initial technical service account.
This token does not have to support all API endpoints (that could depend on an actual user being present). However, it should at least support creating new "real users"  with username/password authentication (in our case https://github.com/mamba-org/quetz-sql-authenticator).

# Changes

Introduce a new config element `users.supertoken` and allow the `Rules` class to bypass this token in the server role assertions.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206579607918088

### DIFF
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -70,7 +70,7 @@ Configure default user permissions, creating default channel and super-admin per
    [users]
    # an optional supertoken that can be used to bypass authorization
    # e.g. for CI pipelines that need to rely on pre-configured tokens for the initial
-   # technial superuser. Length should be at least 32 characters.
+   # technial superuser. Length must be at least 32 characters.
    supertoken = "use `openssl rand -hex 32` to generate a random token"
    # users with owner role
    admins = ["github:admin_user"]
--- a/docs/source/deploying/configuration.rst
+++ b/docs/source/deploying/configuration.rst
@@ -68,6 +68,10 @@ Configure default user permissions, creating default channel and super-admin per
 .. code::
 
    [users]
+   # an optional supertoken that can be used to bypass authorization
+   # e.g. for CI pipelines that need to rely on pre-configured tokens for the initial
+   # technial superuser. Length should be at least 32 characters.
+   supertoken = "use `openssl rand -hex 32` to generate a random token"
    # users with owner role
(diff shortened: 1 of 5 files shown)


docs/source/deploying/configuration.rst
```diff
@@ -68,6 +68,10 @@ Configure default user permissions, creating default channel and super-admin per
 .. code::
 
    [users]
+   # an optional supertoken that can be used to bypass authorization
+   # e.g. for CI pipelines that need to rely on pre-configured tokens for the initial
+   # technial superuser. Length should be at least 32 characters.
```
janjagusch, line 73: You validate the length in the constructor.

```suggestion
   # technial superuser. Length must be at least 32 characters.
```

quetz/authorization.py
```diff
@@ -34,8 +34,37 @@ class ServerRole(str, enum.Enum):
...
+            Supertoken that can be used to bypass role checks
+        """
+        if supertoken and API_key == supertoken:
+            if len(supertoken) < 32:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail="Supertoken too short, must be at least 32 characters long",
+                )
+            self.API_key = None
+            self.is_supertoken = True
```
janjagusch, line 63: can we mark this attribute as private and expose it through a property?

janjagusch (approved):

quetz/authorization.py
```diff
@@ -135,6 +168,9 @@ def assert_assign_user_role(self, role: Optional[str]):
             return self.assert_server_roles([SERVER_OWNER, SERVER_MAINTAINER])
 
     def assert_server_roles(self, roles: list, msg: Optional[str] = None):
+        if self.is_supertoken:
+            return "supertoken"
```
janjagusch, line 172: I don't like how we're sneaking the super token into the authorization layer here. But I also understand that anything more explicit would require big changes in the Quetz authentication and authorization model.

PR #695, opened 2024-04-26
fix: Patch `dummylogin` vulnerability

robinholzi:
# Motivation

The quetz API offers  a`dummy_login` endpoint for development purposes. This endpoint allows a developer to bypass the authentication layer.

Unfortunately, quetz doesn't take any measure to prevent an unauthorized third party from using that endpoint in production.

# Changes

To address this critical vulnerability, I propose to to only expose this endpoint to the API conditionally. For that, it makes sense to use the existing `dev` flag. To pass that into the API subprocess it is convenient to use a file configuration element instead of a cli flag. Therefore I propose to remove the cli `dev` flag and replace it with the config file equivalent.


robinholzi:
@janjagusch I resolved all test failure now (except from `test_download_remote_file_in_parallel`). Locally this test is passing. Using the `create-tables` CI matrix variant it also passes in the CI, `use-migrations` makes problems though. As this test also failed on [main](https://github.com/mamba-org/quetz/actions/runs/8844981415/job/24287954292) I assume that it's not related to my PR.

PR #705, opened 2024-09-03
fix: Include '.conda' packages for reindexing from package store

### LINKED ISSUE #703: Allow reindexing `.conda` packages
We recently had to reindex our package store and noticed only `tar.bz2` packages were showing up after reindexing.

From the code it seems only `.tar.bz2` files are considered during reindex:
https://github.com/mamba-org/quetz/blob/56ab2cfc58c89111637c62f9fc54bec8c97f522e/quetz/tasks/reindexing.py#L140

Can we also enable it for `.conda` packages?

najose:
Currently only `.tar.bz2` packages are considered for reindexing. This PR will allow reindexing to see `.conda` packages too.

fixes #703

### DIFF
--- a/plugins/quetz_mamba_solve/conda-requirements.txt
+++ b/plugins/quetz_mamba_solve/conda-requirements.txt
@@ -1,2 +1,2 @@
 mamba
-conda-build >=3.20
+conda-build >=3.20,<24.7.0
--- a/quetz/tasks/reindexing.py
+++ b/quetz/tasks/reindexing.py
@@ -137,7 +137,7 @@ def reindex_packages_from_store(
     user_id = uuid_to_bytes(user_id)
     pkgstore = config.get_package_store()
     all_files = pkgstore.list_files(channel_name)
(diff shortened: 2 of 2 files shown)


najose:
@wolfv, thanks for the commit to fix the CI, seems like some of the workflows are still failing.

kuepe-sl:
I recently started trying out the new `.conda` packages and they work mostly well - except that they don't appear as "current version" in the Quetz API.

I tracked the bug down to the missing reindexing, which causes multiple packages to have `version_order = 0` in the database.

It would be really awesome if you could merge this, so that the issue is solved, because currently the API is pretty broken as soon as `.conda` packages are involved.

najose:
@wolfv, Can we merge this change? Are the CI errors related to the change in the PR?

kuepe-sl:
@wolfv Any news on this?

The ".conda" package format is now the defauilt with Conda Build, thus it will break Quetz by default now.

faameunier:
@wolfv Any chance we can get this merged? Thanks!

Issue #699, opened 2024-06-03
Not installable with strict priority? LibMambaUnsatisfiableError: Encountered problems while solving


pmlandwehr:
I've been trying to get quetz working with a base miniconda environment, but am getting errors during the basic setup when I have a strict environment required. This might be due to my installing with conda and not mamba, but since the two should be functionally identical when using libmamba as the solver it seems kind of mysterious. (Perhaps this is just a "bug" in that the README could be clarified for folks like myself.)

conda info:
```
     active environment : base
    active env location : /home/pmlandwehr/anaconda3
            shell level : 1
       user config file : /home/pmlandwehr/.condarc
 populated config files : /home/pmlandwehr/.condarc
          conda version : 24.5.0
    conda-build version : not installed
         python version : 3.12.3.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=sandybridge
                          __conda=24.5.0=0
                          __cuda=11.4=0
                          __glibc=2.35=0
                          __linux=6.5.0=0
                          __unix=0=0
       base environment : /home/pmlandwehr/anaconda3  (writable)
      conda av data dir : /home/pmlandwehr/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/pmlandwehr/anaconda3/pkgs
                          /home/pmlandwehr/.conda/pkgs
       envs directories : /home/pmlandwehr/anaconda3/envs
                          /home/pmlandwehr/.conda/envs
               platform : linux-64
             user-agent : conda/24.5.0 requests/2.31.0 CPython/3.12.3 Linux/6.5.0-35-generic ubuntu/22.04.4 glibc/2.35 solver/libmamba conda-libmamba-solver/24.1.0 libmambapy/1.5.8 aau/0.4.4 c/. s/. e/.
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False
```

My .condarc:
```
channels:
  - conda-forge
  - defaults
channel_priority: strict
```
In this case, I get this error when running:
```
$ conda env create -f environment.yml
Channels:
[...]

PR #712, opened 2025-03-18
Limit python version and depend on hatchling

hmaarrfk:


environment.yml
```diff
@@ -2,8 +2,9 @@ name: quetz
 channels:
   - https://repo.mamba.pm/conda-forge
 dependencies:
-  - python>=3.7
+  - python>=3.7,<3.12.0a
```
hmaarrfk, line 5: See https://github.com/mamba-org/quetz/issues/699

